PR #12, opened 2022-05-17
DOPS-1741: Deploy Iroha2 Block Explorer testing env

BAStos525:
# Task
[DOPS-1741]: Deploy Iroha2 Block Explorer testing env

## Changes
Switch to unprevilidge nginx container and update Dockerfile according to it

## Author
Sign-by-off: Vasilii Zyabkin <zyabkin@soramitsu.co.jp>

[DOPS-1741]: https://soramitsu.atlassian.net/browse/DOPS-1741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,7 @@
-FROM node:latest as build-stage
-WORKDIR /app
-COPY package*.json ./
-RUN npm install
-COPY ./ .
-RUN npm run build
-
-FROM nginxinc/nginx-unprivileged:mainline
-USER nginx
-COPY --from=build-stage /app/dist /usr/share/nginx/html
+FROM      node:18.2.0-bullseye
+COPY      . /app
+WORKDIR   /app
+RUN       npm install && npm run build
+USER      node
+ENV       PORT=8080
+CMD       ["node", "app"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,3 @@
-FROM      node:18.2.0-bullseye
-COPY      . /app
-WORKDIR   /app
-RUN       npm install && npm run build
-USER      node
-ENV       PORT=8080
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -6,6 +6,8 @@ COPY ./ .
 RUN npm run build
 
 FROM nginxinc/nginx-unprivileged:mainline
-RUN mkdir /app
-COPY --from=build-stage /app/dist /app
-COPY nginx.conf /etc/nginx/nginx.conf
+ENV LOAD_DIR=/app
+RUN adduser --disabled-password --gecos "" iroha && \
+    mkdir -p ${LOAD_DIR} && \
+    chown -R iroha ${LOAD_DIR}
```
wpi86, line 12: Best practice is not to do this, as executables should be immutable to reduce the attack surface. I think you should give minimum sufficient permissions

wpi86 (changes requested):

Dockerfile
```diff
@@ -6,6 +6,10 @@ COPY ./ .
 RUN npm run build
 
 FROM nginxinc/nginx-unprivileged:mainline
-RUN mkdir /app
-COPY --from=build-stage /app/dist /app
-COPY nginx.conf /etc/nginx/nginx.conf
+ENV LOAD_DIR=/app
+USER root
+RUN mkdir -p ${LOAD_DIR} && \
```
Cre-eD, line 11: I see no sense, you create a folder as root, but then change permissions

Cre-eD (commented):
don't use `node:latest` specify version

Cre-eD (approved):

PR #15, opened 2022-08-01
Feature/main-page-data-and-other-changes

antyas:


src/lib/time.ts
```diff
@@ -0,0 +1,33 @@
...
+export const elapsed = {
+  allMinutes: getAllElapsedMinutes,
+  seconds: getElapsedSeconds,
+};
+
+function formatXX(item: number) {
+  return item < 10 ? '0' + item : item;
+}
+
+export function format(timestamp: string) {
```
0x009922, line 24: I'd recommend to define an [opaque](https://github.com/sindresorhus/type-fest/blob/main/source/opaque.d.ts) `Timestamp` type to make types stronger:

```ts
type Timestamp = Opaque<string, 'ISO timestamp'>
```

src/views/AssetsList.vue
```diff
@@ -0,0 +1,101 @@
...
+  <BaseContentBlock :title="$t('assets')" class="assets-list-page">
+    <BaseTable
+      :loading="table.loading.value"
+      :pagination="table.pagination"
+      :items="table.items.value"
+      @next-page="table.nextPage()"
+      @prev-page="table.prevPage()"
+      @set-page="table.setPage($event)"
+      @set-size="table.setSize($event)"
+    >
```
0x009922, line 11: You can refactor `BaseTable` to accept just what `useTable()` returns so you don't need to configure all these props:

```html
<BaseTable :table="table" />
```

src/views/BlocksList.vue
```diff
@@ -15,7 +15,60 @@
...
-          <span class="h-sm cell">{{ $t('validators') }}</span>
+        </div>
+      </template>
+
+      <template #row="{ item }: { item: Block }">
+        <div class="blocks-list-page__row">
+          <span class="cell">
+            <a :href="`/blocks/${item.height}`" class="primary-link">
+              {{ item.height }}
+            </a>
```
0x009922, line 26: Wait, why not `<RouterLink>`? It's not an SPA pattern to reload page on navigation.

src/views/BlocksList.vue
```diff
@@ -15,7 +15,60 @@
...
+
+          <div class="cell">
+            <time class="row-text">{{ format(item.timestamp) }}</time>
+          </div>
+
+          <span class="cell">
+            <BaseCopyRow :name="$t('hash')" :value="item.block_hash">
+              <a :href="`/blocks/${item.height}`" class="primary-link">
+                {{ item.block_hash }}
+              </a>
```
0x009922, line 37: Expect `<RouterLink>` here as well

src/views/BlocksList.vue
```diff
@@ -15,7 +15,60 @@
...
+        </div>
+      </template>
+
+      <template #mobile-card="{ item }: { item: Block }">
+        <div class="blocks-list-page__mobile-card">
+          <div class="blocks-list-page__mobile-row">
+            <span class="h-sm blocks-list-page__mobile-label">{{ $t('height') }}</span>
+            <a :href="`/blocks/${item.height}`" class="primary-link">
+              {{ item.height }}
+            </a>
```
0x009922, line 51: Expect `<RouterLink>` here as well

src/views/BlocksList.vue
```diff
@@ -15,7 +15,60 @@
...
+            <span class="h-sm blocks-list-page__mobile-label">{{ $t('age') }}</span>
+            <time class="row-text">{{ format(item.timestamp) }}</time>
+          </div>
+
+          <div class="blocks-list-page__mobile-row">
+            <span class="h-sm blocks-list-page__mobile-label">{{ $t('hash') }}</span>
+            <BaseCopyRow :name="$t('hash')" :value="item.block_hash">
+              <a :href="`/blocks/${item.height}`" class="primary-link">
+                <ShortHash :hash="item.block_hash" />
+              </a>
```
0x009922, line 64: Expect `<RouterLink>` here as well

src/views/DomainsList.vue
```diff
@@ -11,9 +11,52 @@
...
+          <span class="h-sm cell">{{ $t('accounts') }}</span>
+        </div>
+      </template>
+
+      <template #row="{ item }: { item: Domain }">
+        <div class="domains-list-page__row">
+          <span class="cell">
+            <a :href="`/domains/${item.id}`" class="primary-link">
+              {{ item.id }}
+            </a>
```
0x009922, line 26: Expect `<RouterLink>` here as well

src/views/DomainsList.vue
```diff
@@ -11,9 +11,52 @@
...
+        </div>
+      </template>
+
+      <template #mobile-card="{ item }: { item: Domain }">
+        <div class="domains-list-page__mobile-card">
+          <div class="domains-list-page__mobile-row">
+            <span class="h-sm domains-list-page__mobile-label">{{ $t('name') }}</span>
+            <a :href="`/domains/${item.id}`" class="primary-link">
+              {{ item.id }}
+            </a>
```
0x009922, line 43: Expect `<RouterLink>` here as well

0x009922 (changes requested):
LGTM, expect the moment of using raw `<a>` instead of `<RouterLink>`

src/views/AssetsList.vue
antyas: I left the possibility for custom logic for tables. but probably in this project such freedom is not required. I will redo it in the next pr

src/lib/time.ts
antyas: I do not think that this will add reliability, but I probably will use it later

src/views/BlocksList.vue
antyas: I will definitely fix this in future pr. I'm shocked how could I let this happen.

0x009922 (approved):

PR #16, opened 2022-08-05
Feature/fsd

antyas:


vite.config.ts
```diff
@@ -18,8 +18,15 @@ export default defineConfig({
...
-      { find: 'styles', replacement: resolve(__dirname, './src/styles/main.scss') },
+      { find: 'styles', replacement: resolve(__dirname, './src/shared/ui/styles/main.scss') },
+      { find: '~app', replacement: resolve(__dirname, './src/app') },
+      { find: '~pages', replacement: resolve(__dirname, './src/pages') },
+      { find: '~widgets', replacement: resolve(__dirname, './src/widgets') },
+      { find: '~features', replacement: resolve(__dirname, './src/features') },
+      { find: '~entities', replacement: resolve(__dirname, './src/entities') },
+      { find: '~shared', replacement: resolve(__dirname, './src/shared') },
+      { find: '~base', replacement: resolve(__dirname, './src/shared/ui/components') },
+      { find: '~icons', replacement: resolve(__dirname, './src/shared/ui/icons') },
```
0x009922, line 29: I'd define a shorthand such as

```ts
const resolveSrc = (...paths) => resolve(__dirname, './src', ...paths)
```

src/entities/account.ts
```diff
@@ -1,9 +1,14 @@
...
-export function countNFTs(account: Account): number {
+function countNFTs(account: Account): number {
   return account.assets
     .reduce((sum, a) => a.value.t === 'Store' ? sum + 1 : sum, 0);
 }
+
+export const accountModel = {
+  countCryptos,
+  countNFTs,
+};
```
0x009922, line 14: Not the best solution in terms of tree-shaking. It would be *a bit* more efficient to export function separately.

src/entities/domain.ts
```diff
@@ -1,9 +1,14 @@
...
-export function countNFTs(domain: Domain): number {
+function countNFTs(domain: Domain): number {
   return domain.asset_definitions
     .reduce((sum, ad) => ad.value_type === 'Store' ? sum + 1 : sum, 0);
 }
+
+export const domainModel = {
+  countCryptos,
+  countNFTs,
+};
```
0x009922, line 14: Same as for `./accounts`

src/entities/transaction.ts
```diff
@@ -38,3 +38,7 @@ export function mapTransaction(transaction: TransactionDto): Transaction {
     },
   };
 };
+
+export const transactionModel = {
+  mapFromDto,
+};
```
0x009922, line 44: Same as for `./accounts`

0x009922 (commented):
I am new for FSD, and it looks very interesting for me.

However, I've noticed its violation here, as well as FSD's weak spot that concludes from it. The point is in the shared layer.

FSD guide says:

> Shared — reusable functionality, detached from the specifics of the project/business.

And actually in the `shared` dir I see the only 2 modules that follow this - they are `lib/time` and `modules/table`. They have utilities that don't contain any application-specific logic.

Other parts really confuses me:

- `api` - application DTOs and endpoints. It is the most domain-specific logic I can imagine in the application. It is actually the core of the application, it's core data model the application operates with. And I wonder, where it should be moved within FSD scope? The most close layer I see is Entities layer, but I can't really say that `api` belongs to it. `dto`s - maybe Entities, yes. `http` - implementation details, maybe Entities again, but for me it seems that Entities layer should be a really pure layer without such complex things like HTTP requests. I am confused and I don't know where is the place for `api` in FSD scope. Btw, in DDD `dto` may be placed at Domains layer, and `http` - at Services layer.
- `lib/localization` - nuff said. Again, its really about application specifics and not about something "detached from the specifics of the project/business". And again, I don't know where it should be.
- `model`
  - `header-portal` - App header micro-store. It's a global UI-related state. It is far-far away from "reusable functionality, detached from the specifics of the project/business"
  - `notifications` - same as above
  - `table` - yeah, it looks like a pure shared utility.
- `ui` - base application components, icons & styles. The core word is **application**. I am confused - it is not about this layer, and I can't tell which layer it belongs to again.

So, FSD looks interesting for me, and your implementation reveals a lot for me, but there is a set of unresolved questions for me.

I don't mind if you proceed on your own without any changes in the project structure, but I'd appreciate if you explain me your vision of the issues I've highlighted.

src/entities/account.ts
antyas: This is one of the necessary sacrifices. Since the logic is divided into modules with strong coupling inside, this does not lead to big problems with the weight of imports.

vite.config.ts
antyas: a good idea. I will do so

antyas:
About shared. something that does not apply to the application at all should already be placed in separate packages. the code in the project is still somehow related to the specific logic of the project. Think of shared as a place for common things that are used on all layers above.

The methodology is still young and it still has some shaky places. In fact, current practices are discussed in their telegram channel, and then they are already entered into the documentation.

Api in shared is their direct recommendation. dto should be there. And the frontend-specific data types that are mapped from something will already be in the entities layer.
Shared should have thin api methods. Logic with response processing, mapping and other specific actions will already be on the layers above. but we only need subtle methods so far.

the shared should also have a basic ui without business logic, from which the ui will be built on the remaining layers. I don’t have all the components there clean, but I will fix it in the future, now I just moved it to the right place.

> It is far-far away from "reusable functionality, detached from the specifics of the project/business"

These are project-specific things, but they are auxiliary and do not relate directly to business logic.

I also realized that I made a mistake, and there should not be a model in the shared. Now I will also think about the role of the things that lie there

0x009922 (approved):

PR #17, opened 2022-09-23
last antyas pr

antyas:


src/shared/ui/components/BaseDropdown.vue
```diff
@@ -32,26 +32,26 @@ type DropdownItem = {
 }
 
 type Props = {
-  modelValue: string | number,
+  value: string | number,
```
0x009922, line 35: Why `value` instead of `moduleValue`? The latter one allows for a shorter `v-model` instead of `v-model:value`, and its just a standard Vue 3 way to define a single model value.

src/shared/ui/components/BaseRadioGroup.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <div class="base-radio-group">
+    <div
+      v-for="(item, i) in props.items"
+      :key="i"
+      class="base-radio-group__item"
+      :data-active="item.value === props.value || null"
```
0x009922, line 7: `data-` attributes for boolean switches, again.

src/shared/ui/components/BaseTable.vue
```diff
@@ -101,22 +108,22 @@ const segmentInfo = computed(() => {
 const numbers = computed(() => {
   const isMobile = width.value < PAGINATION_BREAKPOINT;
   const max = isMobile ? 6 : 10;
-  const side = isMobile ? 4 : 8;
+  const side = isMobile ? 4 : 7;
   const offset = isMobile ? 1 : 3;
```
0x009922, line 112: Too much magic constants.

src/shared/ui/components/BaseTabs.vue
```diff
@@ -20,11 +20,11 @@ type TabItem = {
 
 type Props = {
   items: TabItem[],
-  modelValue: string,
+  value: string,
 }
```
0x009922, line 24: Same as for `<BaseDropdown>` - why do you change `modelValue` to `value`?

0x009922 (changes requested):
The most critical change I request - change inline mocked API to Cypress-based workflow. It has a great support for [intercepting requests and stubbing their responses with fixtures](https://docs.cypress.io/api/commands/intercept#With-a-fixture). 

Using Cypress will lead to:

- a better DX
- a cleaner and lighter code without tree-shaking problems

src/shared/api/fake-http.ts
```diff
@@ -0,0 +1,69 @@
...
+import { makeDomain } from './mock-factories/domain';
+import { transactionList } from './mock-factories/transaction';
+import { list, pagination } from './mock-factories/utils';
+
+const accounts = list(makeAccount, 200);
+const assets = list(makeAsset, 200);
+const assetDefinitions = list(makeAssetDefinition, 100);
+const domains = list(makeDomain, 50);
+const blocks = list(makeBlock, 50);
+const transactions = transactionList(400);
```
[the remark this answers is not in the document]
antyas: The approach is designed for the fact that with a non-working backend, the application should simulate full performance. the same source of mocks can in theory be used in cypress. But in the opposite direction, the inclusion of cypress for the sake of development independent of the backend looks redundant.

src/shared/api/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  fetchPeers(): Promise<Peer[]>;
+  fetchPeerStatus(): Promise<Status>;
+  fetchRoles(): Promise<Role[]>;
+  fetchBlocks(params?: PaginationParams): Promise<Paginated<BlockShallow>>;
+  fetchTransactions(params?: PaginationParams): Promise<Paginated<TransactionDto>>;
+}
+
+export const http: HTTPService = import.meta.env.VITE_FAKE_API_ENABLED === 'TRUE'
+  ? await import('./fake-http')
+  : await import('./http');
```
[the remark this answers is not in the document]
antyas: This is a dynamic import, loading will be at runtime. about the same as the pages in the router

src/shared/ui/components/BaseDropdown.vue
antyas: moduleValue is not best practice, and is more for those who are used to vue 2.
This view gives us more consistency if we need to add additional v-models.

src/shared/ui/components/BaseLink.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+  <router-link :to="to" class="base-link" :data-monospace="monospace || null">
```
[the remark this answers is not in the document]
antyas: constructions with classes are more cumbersome in the template. date attributes have not had performance problems for a very long time, as they used to. There are even css methodologies tied to their active use.
I came to the conclusion that modifiers made this way are more obvious in the template, more flexible and convenient in css

src/pages/TransactionsList.vue
```diff
@@ -0,0 +1,168 @@
...
+} from '~features/filter-transactions';
+
+const HASH_BREAKPOINT = 1200;
+
+const status = ref<ftm.Status>(null);
+const tab = ref<ftm.Tab>('all');
+
+const { width } = useWindowSize();
+
+const hashType = computed(() => width.value < HASH_BREAKPOINT ? 'short' : 'full');
```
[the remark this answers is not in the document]
antyas: I think you're right

src/shared/ui/components/BaseTable.vue
antyas: It seemed to me that from the context of the code it is clear where and what these values are used for. but you can rewrite the algorithm as you like

src/shared/ui/components/BaseDropdown.vue
0x009922: `modelValue` is a **default** prop for `v-model`. It is not a best practice, neither is yours. Your solution is reasonable **if there are other models**. In fact, there are 0 such cases.

src/shared/ui/components/BaseLink.vue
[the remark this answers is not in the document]
0x009922: Yeah, fine. Anyway, waiting for [`.bool` modifier](https://github.com/leopiccionia/rfcs/blob/leopiccionia/bool-modifier/active-rfcs/0000-v-bind-bool.md).

src/shared/ui/components/BaseTable.vue
0x009922: It has more cognitive pressure than reading of something like:

```ts
const MOBILE_SIDE_SIZE = 4
const DESKTOP_SIDE_SIZE = 7

const side = isMobile ? MOBILE_SIDE_SIZE : DESKTOP_SIDE_SIZE
```

.env.local.example
```diff
@@ -1 +1,2 @@
-VITE_API_URL=http://localhost:4000/api/v1
\ No newline at end of file
+VITE_API_URL=http://localhost:4000/api/v1
+VITE_FAKE_API_ENABLED=FALSE
```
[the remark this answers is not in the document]
0x009922: If you can document your code with types + get intellisence hints, why not?

0x009922 (approved):

PR #18, opened 2022-10-13
[refactor]: do revision

0x009922:
- use `pnpm`
- not sure, but I guess type-check wasn't working; now works for sure anyway
- setup `lint` script
- fix deps
- extend `ImportMetaEnv` type
- use `modelValue` for default `v-model`; `useVModel()`
- chores in ESLint & TypeScript configs
- use `computedEager` to improve performance
- use `block.height` as a key at `<LatestBlocks>`


6r1d (approved):
LGTM

PR #19, opened 2022-10-17
Add a store key count in an asset

6r1d:
Adds a template of the `Store` asset key count as discussed on the meeting. Modifies only the mocks.


0x009922 (commented):
It isn't a good solution, imo.

This is not a place to "render" values. And actually the code does the right thing - it returns an empty metadata object for `Store` asset.

Rendering should be changed at the **view** layer.

PR #20, opened 2022-10-19
[feature]: render `Store` asset as a count of keys

0x009922:
Also make `AssetValue` type stronger
and refactor mocks to be type-stronger as well.


6r1d (approved):
Understood, #19 did not follow [FSD](https://feature-sliced.design/) in that regard. Thanks, LGTM.

6r1d (approved):

PR #24, opened 2022-11-22
[refactor] Remove URL shortening

6r1d:
Demonstrates what happens if we remove shortening on the Latest Transactions.
I believe the alternative solution with https://github.com/soramitsu/iroha2-block-explorer-web/pull/23 is better.

Screenshot_2022-11-22_08-03-37


0x009922 (commented):
I believe that with an increased `line-height` text wrapping for hash will look much better. Consider asking a designer about which `line-height` exactly is suitable.

0x009922:
Not relevant anymore.

PR #27, opened 2023-10-20
[feature] : Add BlockDetails Page

horizenight:
## Description

Detailed comments:
Adding the BlockDetails functionality

### DIFF
--- a/src/entities/transaction/model.ts
+++ b/src/entities/transaction/model.ts
@@ -14,7 +14,7 @@ declare global {
       time_to_live_ms: number;
       nonce: null | number;
       metadata: any;
-    }
+    };
     signatures: Signature[];
     rejection_reason?: string;
   }
@@ -48,6 +48,15 @@ export function mapFromDto(transaction: TransactionDto): Transaction {
 
 export async function fetchList(params?: PaginationParams): Promise<Paginated<Transaction>> {
   const res = await http.fetchTransactions(params);
+  const data = res.data.map(mapFromDto);
+
+  return {
+    pagination: res.pagination,
+    data,
+  };
+}
+export async function fetchBlocksList(selectedTransaction: TransactionDto[], params?: PaginationParams): Promise<Paginated<Transaction>> {
+  const res = await http.fetchBlockTransactions(selectedTransaction);
   const data = res.data.map(mapFromDto);
 
   return {
--- a/src/entities/transaction/model.ts
+++ b/src/entities/transaction/model.ts
@@ -57,7 +57,7 @@ export async function fetchList(params?: PaginationParams): Promise<Paginated<Tr
   };
 }
 // TODO  paginated List => make generic change a list to paginated list
-export async function fetchBlocksList(selectedTransaction: TransactionDto[], params?: PaginationParams): Promise<Paginated<Transaction>> {
+export function fetchBlocksList(selectedTransaction: TransactionDto[], params?: PaginationParams): Paginated<Transaction> {
   const res = pagination(selectedTransaction, params);
   const data = res.data.map(mapFromDto);
   return {
--- a/src/entities/transaction/model.ts
+++ b/src/entities/transaction/model.ts
@@ -1,5 +1,6 @@
 import { Instruction } from '@iroha2/data-model';
 import { http } from '~shared/api';
+import { pagination } from '~shared/api/mock-factories/utils';
 
 declare global {
   export interface Transaction {
@@ -55,10 +56,10 @@ export async function fetchList(params?: PaginationParams): Promise<Paginated<Tr
     data,
   };
 }
+// TODO  paginated List => make generic change a list to paginated list
 export async function fetchBlocksList(selectedTransaction: TransactionDto[], params?: PaginationParams): Promise<Paginated<Transaction>> {
-  const res = await http.fetchBlockTransactions(selectedTransaction);
+  const res = pagination(selectedTransaction, params);
   const data = res.data.map(mapFromDto);
-
   return {
     pagination: res.pagination,
     data,
--- a/src/features/blocks/utils/useBlockTable.ts
+++ b/src/features/blocks/utils/useBlockTable.ts
@@ -1,4 +1,4 @@
-import { Ref, ref, computed } from 'vue';
+import { Ref, ref } from 'vue';
 import { transactionModel } from '~entities/transaction';
 
 export type TableFetchFn<T> = (params: PaginationParams) => Promise<Paginated<T>>;
@@ -13,46 +13,48 @@ export type TablePagination = {
 export function useBlockTable() {
   const loading = ref(false);
   const selectedTransaction: Ref<TransactionDto[]> = ref([]);
+  const items: Ref<Transaction[]> = ref([]);
   const pagination: Ref<TablePagination> = ref({
     page: 1,
     page_size: 10,
     pages: 1,
     total: 10,
   });
-
-  const items = computed(() => {
+  function fetch() {
     const res = transactionModel.fetchBlocksList(selectedTransaction.value, {
       page: pagination.value.page,
       page_size: pagination.value.page_size,
     });
-
+    items.value = res.data;
     pagination.value = {
       ...res.pagination,
       pages: Math.ceil(res.pagination.total / res.pagination.page_size),
     };
-
-    return res.data;
-  });
+  }
 
   function setTransaction(transaction: TransactionDto[]) {
     selectedTransaction.value = transaction;
+    fetch();
   }
 
   function nextPage() {
     if (pagination.value.page < pagination.value.pages) {
       pagination.value.page += 1;
+      fetch();
     }
   }
 
   function prevPage() {
     if (pagination.value.page > 1) {
       pagination.value.page -= 1;
+      fetch();
     }
   }
 
   function setPage(index: number) {
     if (index !== pagination.value.page) {
       pagination.value.page = index;
+      fetch();
     }
   }
 
@@ -60,6 +62,7 @@ export function useBlockTable() {
     if (n !== pagination.value.page_size) {
       pagination.value.page_size = n;
       pagination.value.page = 1;
+      fetch();
     }
   }
 
--- a/src/features/blocks/utils/useBlockTable.ts
+++ b/src/features/blocks/utils/useBlockTable.ts
@@ -1,0 +1,76 @@
+import { Ref, ref, computed } from 'vue';
+import { transactionModel } from '~entities/transaction';
+
+export type TableFetchFn<T> = (params: PaginationParams) => Promise<Paginated<T>>;
+
+export type TablePagination = {
+  page: number;
+  page_size: number;
+  pages: number;
+  total: number;
+};
+
+export function useBlockTable() {
+  const loading = ref(false);
+  const selectedTransaction: Ref<TransactionDto[]> = ref([]);
+  const pagination: Ref<TablePagination> = ref({
+    page: 1,
+    page_size: 10,
+    pages: 1,
+    total: 10,
+  });
+
+  const items = computed(() => {
+    const res = transactionModel.fetchBlocksList(selectedTransaction.value, {
+      page: pagination.value.page,
+      page_size: pagination.value.page_size,
+    });
+
+    pagination.value = {
+      ...res.pagination,
+      pages: Math.ceil(res.pagination.total / res.pagination.page_size),
+    };
+
+    return res.data;
+  });
+
+  function setTransaction(transaction: TransactionDto[]) {
+    selectedTransaction.value = transaction;
+  }
+
+  function nextPage() {
+    if (pagination.value.page < pagination.value.pages) {
+      pagination.value.page += 1;
+    }
+  }
+
+  function prevPage() {
+    if (pagination.value.page > 1) {
+      pagination.value.page -= 1;
+    }
+  }
+
+  function setPage(index: number) {
+    if (index !== pagination.value.page) {
+      pagination.value.page = index;
+    }
+  }
+
+  function setSize(n: number) {
+    if (n !== pagination.value.page_size) {
+      pagination.value.page_size = n;
+      pagination.value.page = 1;
+    }
+  }
+
+  return {
+    loading,
+    items,
+    pagination,
+    nextPage,
+    prevPage,
+    setPage,
+    setSize,
+    setTransaction,
+  };
+}
--- a/src/pages/BlockDetails.spec.ts
+++ b/src/pages/BlockDetails.spec.ts
@@ -37,7 +37,14 @@ describe('BlockDetails.vue', async () => {
     expect(blockCardComponent.exists()).toBe(true);
     expect(blockTransactionsListComponent.exists()).toBe(true);
   });
+
 });
+
+describe('BlockDetailsCard.vue', () => {
+  it('renders things fine', () => {
+
+  })
+})
 
 // test2
 // write a test that matches the backend data to mocked data types and if there types matches it passses
--- a/src/pages/BlockDetails.spec.ts
+++ b/src/pages/BlockDetails.spec.ts
@@ -1,0 +1,46 @@
+import { mount, config } from '@vue/test-utils';
+import { describe, it, vi, expect } from 'vitest';
+import BlockDetails from '../pages/BlockDetails.vue';
+import { createRouter, createWebHistory } from 'vue-router';
+import { routes } from '../../src/app/router';
+
+const router = createRouter({
+  history: createWebHistory(),
+  routes,
+});
+
+config.global.mocks = {
+  $t: (msg:string) => msg,
+};
+
+vi.mock('vue-i18n', () => ({
+  useI18n: () => ({
+    t: (msg:string) => msg,
+  }),
+}));
+
+// Test conditional rendering:
+// Verify that elements like the block-details-card and transactions-list-page are only rendered when block has a value.
+describe('BlockDetails.vue', async () => {
+  it('renders the block-details-card and block-transactions-list correctly', async () => {
+    router.push('/blocks/606600');
+    await router.isReady();
+    const wrapper = mount(BlockDetails, {
+      global: {
+        plugins: [router],
+      },
+    });
+    await wrapper.vm.$nextTick();
+    await new Promise(resolve => setTimeout(resolve, 10));
+    const blockCardComponent = wrapper.find('.block-details-card__body');
+    const blockTransactionsListComponent = wrapper.find('.transactions-list-page');
+    expect(blockCardComponent.exists()).toBe(true);
+    expect(blockTransactionsListComponent.exists()).toBe(true);
+  });
+});
+
+// test2
+// write a test that matches the backend data to mocked data types and if there types matches it passses
+
+// test 3: Test pagination:
+// Verify that the pagination controls work as expected, adjusting the displayed transactions based on the selected page.
--- a/src/pages/BlockDetails.vue
+++ b/src/pages/BlockDetails.vue
@@ -20,86 +20,7 @@
 
       <!-- Card Body  -->
       <template #default>
-        <div class="block-details-card__body">
-          <div class="block-details-card__column">
-            <div class="block-details-card__column__row">
-              <span class="h-sm cell">{{
-                `${$t("block", 1)} ${$t("hash")}`
-              }}</span>
-              <BaseHash
-                :hash="block.block_hash"
-                :link="`/blocks/${block.height}`"
-                :type="hashType"
-                copy
-                class="block-details-card__column__row__hash"
-              />
-            </div>
-            <div class="block-details-card__column__row">
-              <span class="h-sm cell">{{
-                $t('blockDetails.transactionMerkleRootHash')
-              }}</span>
-              <BaseHash
-                :hash="block.transactions_merkle_root_hash"
-                :link="`/blocks/${block.height}`"
-                :type="hashType"
-                copy
-                class="block-details-card__column__row__hash"
-              />
-            </div>
-            <div class="blocks-details-card__column__row">
-              <BaseLink
-                :to="`/blocks/${block.height}`"
-                class="block-details-card__column__row__hashFrame"
-              >
-                <!-- TODO wait for design to specify what should happen on click"-->
-                {{
-                  $t("blockDetails.viewChangeProofs", {
-                    length: block.view_change_proofs.length,
-                  })
-                }}
-              </BaseLink>
-            </div>
-          </div>
-          <div class="block-details-card__column">
-            <div class="block-details-card__column__row">
-              <span class="h-sm cell">{{
-                $t("blockDetails.parent") +
-                  " " +
-                  $t("block", 1) +
-                  " " +
-                  $t("hash")
-              }}</span>
-              <BaseHash
-                :hash="block.parent_block_hash"
-                :link="`/blocks/${block.height}`"
-                :type="hashType"
-                copy
-                class="block-details-card__column__row__hashFrame"
-              />
-            </div>
-            <div class="block-details-card__column__row">
-              <span class="h-sm cell">{{
-                $t('blockDetails.rejectedTransactionMerkleRootHash')
-              }}</span>
-              <BaseHash
-                :hash="block.rejected_transactions_merkle_root_hash"
-                :link="`/blocks/${block.height}`"
-                :type="hashType"
-                copy
-                class="block-details-card__column__row__hash"
-              />
-            </div>
-            <div class="blocks-details-card__row">
-              <BaseLink
-                :to="`/blocks/${block.height}`"
-                class="block-details-card__column__row__hashFrame"
-              >
-                <!-- TODO wait for design to specify what should happen on click"-->
-                {{ $t("blockDetails.invalidatedTransactionHash",{length: block.invalidated_blocks_hashes.length}) }}
-              </BaseLink>
-            </div>
-          </div>
-        </div>
+        <BlockDetailsCard rejected_merkle_root_hash="" block-number="" merkle_root_hash="" parent_block_hash="" view_change_proofs="" invalidated_block_hashes="" hash="" />
       </template>
     </BaseContentBlock>
   </div>
@@ -177,6 +98,7 @@ import BaseTable from '~base/BaseTable.vue';
 import BaseHash from '~shared/ui/components/BaseHash.vue';
 import BaseLink from '~shared/ui/components/BaseLink.vue';
 import ArrowIcon from '~icons/arrow.svg';
+import BlockDetailsCard from "~pages/BlockDetailsCard.vue";
 
 const WINDOW_WIDTH_BREAKPOINT_TO_SHRINK_HASH = 1200;
 
--- a/src/pages/BlockDetails.vue
+++ b/src/pages/BlockDetails.vue
@@ -1,0 +1,339 @@
+<template>
+  <div v-if="block && block.height">
+    <BaseContentBlock :title="$t('Block ') + '#' + block.height" class="block-details-card">
+      <!-- Card Header -->
+      <template #header>
+        <div class="block-details-card__title">
+          <ArrowIcon />
+          <h2 class="base-content-block__title">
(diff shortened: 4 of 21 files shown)


src/shared/api/mock-factories/block.ts
```diff
@@ -10,3 +11,17 @@ export function makeBlock(block_hash?: string): BlockShallow {
...
+
+export function makeBlockDetails(height?: number): Block {
+  return {
+    block_hash: hash(64),
+    parent_block_hash: hash(64),
+    rejected_transactions_merkle_root_hash: hash(64),
+    invalidated_blocks_hashes: [],
+    height: height ?? randNumber(),
+    rejected_transactions: [],
+    timestamp: new Date(Date.now() - 10000).toISOString(),
```
0x009922, line 23: Suggest using date utils from `@ngneat/falso`, e.g. [`randPastDate`](https://ngneat.github.io/falso/docs/date/#randpastdate)

src/shared/lib/table.ts
```diff
@@ -78,3 +79,67 @@ export function useTable<T>(fetchFn: TableFetchFn<T>) {
...
+    loading.value = true;
+    const res = await transactionModel.fetchBlocksList(selectedTransaction, {
+      page: pagination.page,
+      page_size: pagination.page_size,
+    });
+    items.value = res.data;
+    pagination.page = res.pagination.page;
+    pagination.page_size = res.pagination.page_size;
+    pagination.total = res.pagination.total;
+    pagination.pages = Math.ceil(res.pagination.total / res.pagination.page_size);
```
0x009922, line 103: You can define `pagination` with `ref()` instead of `reactive` and replace the whole object at once:

```ts
const pagination = ref<null | TablePagination>(null)

// ...

pagination.value = {
  ...res.pagination,
  pages: Math.ceil(res.pagination.total / res.pagination.page_size)
}
```

src/shared/api/index.ts
```diff
@@ -10,6 +11,8 @@ interface HTTPService {
   fetchPeerStatus(): Promise<Status>;
   fetchRoles(): Promise<Role[]>;
   fetchBlocks(params?: PaginationParams): Promise<Paginated<BlockShallow>>;
+  fetchBlocksDetails(id: number): Promise<Block>;
+  fetchBlockTransactions(transactions:TransactionDto[], params?: PaginationParams):Promise<Paginated<TransactionDto>>;
```
0x009922, line 15: `fetchBlockTransactions` - what API endpoint does it reflect?

This interface is a convenient wrap around backend's API. AFAIK, backend doesn't have a method to get transactions from a certain block.

src/shared/api/dto.ts
```diff
@@ -128,7 +128,7 @@ declare global {
     /**
      * List of serialized {@link @iroha2/data-model#VersionedValidTransaction}
      */
-    transactions: string[];
+    transactions: TransactionDto[];
```
0x009922, line 131: ..why?

Look at the `BlockDTO` at the backend:

https://github.com/soramitsu/iroha2-block-explorer-backend/blob/57da68535e2043e0694ae4049ffac83bae2565d1/src/web/blocks.rs#L64

It has `transactions: Vec<SerScaleHex<TransactionValue>>`. Thus, in JSON it will be `string[]`.

src/shared/api/http.ts
```diff
@@ -55,6 +55,10 @@ export function fetchBlocks(params?: PaginationParams): Promise<Paginated<BlockS
   return get('/blocks', params);
 }
 
+export function fetchBlocksDetails(block_hash: string): Promise<Block> {
```
0x009922, line 58: Look at the [API reference](https://github.com/soramitsu/iroha2-block-explorer-backend/blob/57da68535e2043e0694ae4049ffac83bae2565d1/api.md#get-apiv1blocksheight-or-hash):

> `/api/v1/blocks/{height or hash}`

Therefore, to make `fetchBlockDetails` more accurate, it should be:

```suggestion
export function fetchBlocksDetails(height_or_hash: string | number): Promise<Block> {
```

src/shared/lib/table.ts
```diff
@@ -78,3 +79,67 @@ export function useTable<T>(fetchFn: TableFetchFn<T>) {
     setSize,
   };
 }
+// for generating blocksListTable in paginated way
+export function useBlockTable(selectedTransaction: TransactionDto[]) {
```
0x009922, line 83: `shared/lib/table` contains _shared_ utils related to tables all across the application. `useBlockTable` should be placed somewhere in the scope of the block details page/feature. You need to comprehend how [Feature-Sliced Design](https://feature-sliced.design/) works.

Apart from that, `useBlockTable` seems to be almost a full copy of `useTable`.

src/pages/BlockDetails.vue
```diff
@@ -0,0 +1,339 @@
...
+            </div>
+          </div>
+
+          <div class="transactions-list-page__column">
+            <div class="transactions-list-page__label">Block</div>
+
+            <BaseLink :to="'/blocks/' + item.block_height">
+              {{ item.block_height }}
+            </BaseLink>
+          </div>
```
0x009922, line 125: This column is probably redundant. All transactions displayed in this table already belong to the block which displays the table.

src/pages/BlockDetails.vue
```diff
@@ -0,0 +1,339 @@
...
+    ? route.params.id.join('')
+    : route.params.id;
+  const height = parseInt(heightString, 10); // Convert the string to an integer
+  console.log(height);
+  const fetchData = async () => {
+    try {
+      const blockDetails = await http.fetchBlocksDetails(height);
+      block.value = blockDetails;
+      console.log(blockDetails);
+      table = useBlockTable(blockDetails.transactions);
```
0x009922, line 179: You should avoid using composables outside of the scope of the component.

`useBlockTable` will initialize reactive logic on invocation (such as refs, reactive objects, watchers, hooks etc). In order for this reactive logic to work properly (e.g. destroy watchers when the component is disposed), it should be initialised within component's _setup_ stage, or during its lifecycle hooks.

In this case, however, you invoke an async function at any time and then initialize reactive data within `useBlockTable`. Vue will not track the reactivity properly.

src/pages/BlockDetails.vue
```diff
@@ -0,0 +1,339 @@
...
+
+let table = useBlockTable([]);
+const block = ref();
+
+onBeforeMount(() => {
+  // Access the route object and get the 'id' parameter
+  const route = useRoute(); // Use useRoute to access the $route object
+  const heightString = Array.isArray(route.params.id)
+    ? route.params.id.join('')
+    : route.params.id;
```
0x009922, line 171: I guess `params.id` should never be an array. If it is, it is an exception.

src/pages/BlockDetails.vue
```diff
@@ -0,0 +1,339 @@
...
+  TransactionTypeFilter,
+} from '~features/filter-transactions';
+
+import BaseContentBlock from '~base/BaseContentBlock.vue';
+import BaseTable from '~base/BaseTable.vue';
+import BaseHash from '~shared/ui/components/BaseHash.vue';
+import BaseLink from '~shared/ui/components/BaseLink.vue';
+import ArrowIcon from '~icons/arrow.svg';
+
+const HASH_BREAKPOINT = 1200;
```
0x009922, line 152:
```suggestion
const WINDOW_WIDTH_BREAKPOINT_TO_SHRINK_HASH = 1200;
```

0x009922 (changes requested):

src/shared/api/index.ts
```diff
@@ -1,4 +1,5 @@
 interface HTTPService {
+[x: string]: any;
```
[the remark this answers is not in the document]
horizenight: This was overhead added by the ```VScode```. 
Removed it in latest commit.

src/shared/api/mock-factories/block.ts
horizenight: Acknowledged and replaced the timestamp to use ```randPastDate```

src/pages/BlockDetails.vue
```diff
@@ -0,0 +1,339 @@
...
+
+const status = ref<ftm.Status>(null);
+const tab = ref<ftm.Tab>('all');
+
+const { width } = useWindowSize();
+
+const hashType = computedEager(() =>
```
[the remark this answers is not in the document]
horizenight: acknowledged and committed 👍🏼

src/shared/lib/table.ts
horizenight: While I kept most of the code similar to the previous implementation. Using ```ref()``` gives more advantage. 
Acknowledged and committed 👍🏼

src/pages/BlockDetails.vue
horizenight: This column is subject to discussion as per design for the ```BlockDetails Page```.

src/pages/BlockDetails.vue
horizenight: acknowledged and committed 👍🏼

src/pages/BlockDetails.vue
horizenight: This was due to ```params.id``` has type ```string | string[]``` where as ```parseInt()``` function only accepts ```string``` . Better solution defining the type.

```js
const heightString: string = route.params.id as string;
```

src/shared/api/http.ts
horizenight: acknowledged and committed 👍🏼

src/pages/BlockDetails.vue
```diff
@@ -0,0 +1,339 @@
...
+          <div class="blocks-list-page__row">
+            <BaseLink :to="`/blocks/${block.height}`" class="block-details-card__row__hashFrame">
+              {{ $t('6 View Change Proofs') }}
+            </BaseLink>
+          </div>
+          <div class="blocks-list-page__row">
+            <BaseLink :to="`/blocks/${block.height}`" class="block-details-card__row__hashFrame">
+              {{ $t('13 Invalidated Block Hashes') }}
+            </BaseLink>
+          </div>
```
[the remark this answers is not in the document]
horizenight: Acknowledged and Committed 👍🏼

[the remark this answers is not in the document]
horizenight: What should click on ```view_change_proof``` and ```invalidated_block_hashes``` show ?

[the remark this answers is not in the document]
0x009922: The design doesn't specify it. I guess we could do it unclickable for now and leave a comment like "TODO wait for design to specify what should happen on click".

src/pages/BlockDetails.vue
0x009922: Or explicitly validating the type in runtime, which might be safer. For example, using `tiny-invariant` library:

```js
import invariant from 'tiny-invariant'

const height = route.params.id

// after this TypeScript will be sure it is a `string`
invariant(typeof height === 'string')
```

src/pages/BlockDetails.vue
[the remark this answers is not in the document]
horizenight: Acknowledged and Committed 👍🏼

src/shared/api/index.ts
horizenight: At first was using this coupled with the composable to return a ```paginated transactions wrap``` , it no longer serves any purpose.

src/pages/BlockDetails.vue
horizenight: Acknowledged and committed 👍🏼

src/pages/BlockDetails.vue
```diff
@@ -54,20 +61,21 @@
           </div>
           <div class="blocks-list-page__row">
             <BaseLink :to="`/blocks/${block.height}`" class="block-details-card__row__hashFrame">
-              {{ $t('6 View Change Proofs') }}
+              <!-- TODO wait for design to specify what should happen on click"-->
+              {{ block.view_change_proofs.length + $t(' View Change Proofs') }}
```
0x009922, line 65: `vue-i18n` supports some features for interpolation. Read more here:

- [Message Format Syntax](https://vue-i18n.intlify.dev/guide/essentials/syntax.html)
- [Pluralization](https://vue-i18n.intlify.dev/guide/essentials/pluralization.html)

src/shared/lib/table.ts
```diff
@@ -26,44 +26,44 @@ export function useTable<T>(fetchFn: TableFetchFn<T>) {
     loading.value = true;
 
     const res = await fetchFn({
-      page: pagination.page,
-      page_size: pagination.page_size,
+      page: pagination.value.page,
+      page_size: pagination.value.page_size,
     });
```
0x009922, line 31: I think you can just pass the value, i.e. `fetchFn(pagination.value)`

src/shared/lib/table.ts
```diff
@@ -26,44 +26,44 @@ export function useTable<T>(fetchFn: TableFetchFn<T>) {
...
 
     items.value = res.data;
-    pagination.page = res.pagination.page;
-    pagination.page_size = res.pagination.page_size;
-    pagination.total = res.pagination.total;
-    pagination.pages = Math.ceil(res.pagination.total / res.pagination.page_size);
+    pagination.value.page = res.pagination.page;
+    pagination.value.page_size = res.pagination.page_size;
+    pagination.value.total = res.pagination.total;
+    pagination.value.pages = Math.ceil(res.pagination.total / res.pagination.page_size);
```
0x009922, line 37: You can replace `.value` all at once:

```ts
pagination.value = {
  ...res.pagination,
  pages: Math.ceil(res.pagination.total / res.pagination.page_size)
} 
```

src/pages/BlockDetails.vue
```diff
@@ -0,0 +1,335 @@
...
+</template>
+
+<script setup lang="ts">
+import { ref, onMounted } from 'vue';
+import { useRoute } from 'vue-router';
+import { useBlockTable } from '~shared/lib/table';
+import { useWindowSize, computedEager } from '@vueuse/core';
+import { http } from '~shared/api';
+import { TransactionStatus } from '~entities/transaction';
+import { format } from '~shared/lib/time';
```
0x009922, line 139:
```suggestion
import { format as formatTime } from '~shared/lib/time';
```

src/pages/BlockDetails.vue
```diff
@@ -0,0 +1,335 @@
...
+import {
+  filterTransactionsModel as ftm,
```
0x009922, line 141:
```suggestion
  filterTransactionsModel,
```

src/pages/BlockDetails.vue
```diff
@@ -0,0 +1,335 @@
...
+type HashType = 'full' | 'short'
+
+const hashType = computedEager<HashType>(() =>
+  width.value < WINDOW_WIDTH_BREAKPOINT_TO_SHRINK_HASH ? 'short' : 'full',
+);
+
+const route = useRoute();
+const heightString = route.params.id;
+invariant(typeof heightString === 'string');
+const height = parseInt(heightString, 10);
```
0x009922, line 169: Strictly speaking, `route.params.id` is a reactive value. Therefore, `height` should be a computed.

src/shared/api/mock-factories/block.ts
```diff
@@ -1,12 +1,28 @@
...
+  return {
+    block_hash: hash(64),
+    parent_block_hash: hash(64),
+    rejected_transactions_merkle_root_hash: hash(64),
+    transactions_merkle_root_hash: hash(64),
+    invalidated_blocks_hashes: [],
+    height: height ?? randNumber(),
+    rejected_transactions: [],
+    timestamp: randPastDate().toISOString(),
+    transactions: transactionList(400),
```
0x009922, line 25: Unfortunately, it won't work. The actual transactions from the backend will be string HEXes of serialized transactions. You should use `@iroha2/data-model` package in order to work with it.

src/entities/transaction/model.ts
```diff
@@ -55,3 +56,12 @@ export async function fetchList(params?: PaginationParams): Promise<Paginated<Tr
...
 }
+// TODO  paginated List => make generic change a list to paginated list
+export async function fetchBlocksList(selectedTransaction: TransactionDto[], params?: PaginationParams): Promise<Paginated<Transaction>> {
+  const res = pagination(selectedTransaction, params);
+  const data = res.data.map(mapFromDto);
+  return {
+    pagination: res.pagination,
+    data,
+  };
+}
```
0x009922, line 67: Apart from other things I have to say about this function, at the very least it has no reason to be `async` - there is no any `await` inside. Making it `async` just brings complexity (because it will return `Promise<T>` instead of just `T`)

```suggestion
export function fetchBlocksList(selectedTransaction: TransactionDto[], params?: PaginationParams): Paginated<Transaction> {
  const res = pagination(selectedTransaction, params);
  const data = res.data.map(mapFromDto);
  return {
    pagination: res.pagination,
    data,
  };
}
```

src/shared/lib/table.ts
```diff
@@ -78,3 +79,77 @@ export function useTable<T>(fetchFn: TableFetchFn<T>) {
     setSize,
   };
 }
+
+// for generating blocksListTable in paginated way
+export function useBlockTable() {
```
0x009922, line 84: Here are the reasons why I think this composable should be refactored:

- It has duplication of pagination logic in `useTable`
- it has `async function fetch()`, which has no reasons to be async. It calls `await transactionModel.fetchBlocksList`, but this function also has no reasons to be async if you look inside. Since `fetch` is `async`, all other functions also have to be `async`: `setPage`, `setSize` etc.
- `loading` flag is useless, because whenever it is set to `true`, it will be immediately set back to `false`, because there is no any asynchronicity in the actuality.
- the logic of function is kind of messy. Externally, you put data inside through `setTransaction(items)`. Then, in order to get the paginated view of transactions, you call `await fetch()` which takes `selectedTransactions`, applies `pagination.value` to it and sets the value into `items.value`, also updating `pagination.value` for some reason (why?). 
- The name `useBlockTable` suggests that it is responsible for a table of some block. It is located in the `shared/lib/table.ts` next to a generic `useTable` function which works over any kinds of paginated async values. `useBlockTable`, however, belongs to Blocks domain. There are other places in the app structure where such specific logic should be placed.

What I would suggest:

- In order to reduce duplication of reactive logic, try to separate it into a reusable composable
- Whenever you have some data updated, do not manually recompute other data which depends on it. This is exactly what `computed`s and watchers are for. In your case, you manually trigger update of `items` through `fetch()` whenever you update the values `items` depend on (`selectedTransactions`, `pagination`). Instead, you should define `items` as a **`computed`**, which is reactively updated on its own whenever its reactive dependencies change.
- Do not use `async` if you don't have reason for it. `async` is a very useful thing, but it brings redundant complexity if used not properly.

horizenight: acknowledged and committed 👍🏼

src/entities/transaction/model.ts
horizenight: acknowledged and committed 👍🏼

src/pages/BlockDetails.vue
horizenight: Acknowledged and committed 👍🏼
( it would be great if you can point to me a resource where more about these small details can be read. )

src/pages/BlockDetails.vue
horizenight: Acknowledged and committed 👍🏼

src/pages/BlockDetails.vue
horizenight: Acknowledged and committed 👍🏼

src/shared/lib/table.ts
horizenight: Acknowledged and committed 👍🏼

src/pages/BlockDetails.vue
0x009922: [Accessing the Router and current Route inside `setup`](https://router.vuejs.org/guide/advanced/composition-api.html#Accessing-the-Router-and-current-Route-inside-setup)

src/pages/BlockDetails.vue
horizenight: Read , Acknowledged and Committed 👍🏼

src/pages/BlockDetails.vue
```diff
@@ -217,41 +263,43 @@ onMounted(async () => {
...
     }
   }
-
 }
 
 // styles for the blockTable
 .transactions-list-page {
   &__row {
-    width: 100%;
+    width: 10%;
```
0x009922, line 302: Why `10%`?

src/pages/BlockDetails.vue
```diff
@@ -157,9 +205,7 @@ const tab = ref<filterTransactionsModel.Tab>('all');
 
 const { width } = useWindowSize();
 
-type HashType = 'full' | 'short'
-
-const hashType = computedEager<HashType>(() =>
+const hashType = computed<'full' | 'medium' | 'short' | 'two-line'>(() =>
```
0x009922, line 208: It's better to use `computerEager` here, because the value recomputes frequently, recomputation is cheap, and the result is of primitive type.

src/pages/BlockDetails.vue
```diff
@@ -1,12 +1,12 @@
 <template>
   <div v-if="block && block.height">
-    <BaseContentBlock :title="$t('Block ') + '#' + block.height" class="block-details-card">
+    <BaseContentBlock :title="$t('blocks',2) + '#' + block.height" class="block-details-card">
```
0x009922: 1. To use pluralisation feature, you need to use `$tc`, not `$t`.
2. `$t(..) + '#' + block.height` is not how `$t` should be used. It is designed with interpolation in mind. Instead, you should use:
   ```js
   $t('foo.bar', { height: block.height })
   // where `foo.bar` is something like "Block #{height}"

Re-read [syntax](https://vue-i18n.intlify.dev/guide/essentials/syntax.html) and [pluralisation](https://vue-i18n.intlify.dev/guide/essentials/pluralization.html) topics and combine both of my points.

src/shared/lib/localization/en.json
```diff
@@ -6,7 +6,7 @@
   "viewAll": "view all",
   "token": "Token",
   "accounts": "Accounts",
-  "blocks": "Blocks",
+  "blocks": "Blocks | Block",
```
0x009922, line 9: If you use pluralisation (` | `) inside of a message, then the key of the message (`blocks`) here should be in a singular form, i.e. `block` & `Block | Blocks`. Singular comes first here.

0x009922 (changes requested):
Apparently, i18n feature is used incorrectly.

In addition fixing only this, I would welcome if you provide tests for `<BlockDetails>` component, showing how actual content is rendered. I believe currently it simply doesn't work, although I haven't run it.

src/pages/BlockDetails.vue
horizenight: 1. Researched about it `$tc`  is not needed with `composition API` as `$t` supports pluralisation. 
References: [docs](https://arc.net/l/quote/brapfxsi) , [stack overflow](https://arc.net/l/quote/xpiopjne)
2. Acknowledged and committed 👍🏼 
3. For string concatination (Two approaches are their ) : 
Example: 
```
en.json => "parent" : "Parent",
usage=> `${$t("blockDetails.parent")} ${$t("block", 1)} ${$t("hash")}`
or => {{ $t("blockDetails.parent") + " " +$t("blocks", 2) +" " +$t("hash") }}     
output=> Parent Block Hash

```
```
en.json => "parent" : "Parent Block Hash",
usage=>  {{ $t("blockDetails.parent") }} 
output=> Parent Block Hash
```
Kindly , suggest which approach to follow.

src/shared/lib/localization/en.json
horizenight: Acknowledged and committed 👍🏼

src/pages/BlockDetails.vue
horizenight: Acknowledged and committed 👍🏼

src/pages/BlockDetails.vue
0x009922: Try using the following format in messages: `Block #{height}`

src/pages/BlockDetails.vue
```diff
@@ -24,7 +24,7 @@
           <div class="block-details-card__column">
             <div class="block-details-card__column__row">
               <span class="h-sm cell">{{
-                $t("blocks", 2) + " " + $t("hash")
+                `${$t("block", 1)} ${$t("hash")}`
```
0x009922, line 27: This also should be a single `$t` invocation.

src/pages/BlockDetails.vue
```diff
@@ -125,7 +105,7 @@
   </div>
   <!-- Block Transaction Details Component -->
   <BaseContentBlock
-    :title="$t('blocks') + ' ' + $t('transaction')"
+    :title="$t('block',1) + ' ' + $t('transaction',2)"
```
0x009922, line 108: Use a single `$t`

src/shared/lib/localization/en.json
```diff
@@ -53,10 +52,12 @@
...
   "blockDetails":{
+    "title": "Block #{height}",
     "parent" : "Parent",
-    "merkle": "Merkle",
-    "invalidated": "Invalidated",
-    "viewChangeProofs": "View Change Proofs"
+    "transactionMerkleRootHash": "Transaction's Merkle Root Hash",
+    "rejectedTransactionMerkleRootHash": "Rejected Transaction's Merkle Root Hash",
+    "invalidatedTransactionHash": "{length} Invalidated Block Hashes",
+    "viewChangeProofs": "{length} View Change Proofs"
```
0x009922, line 60: What if `length` is 1? It would be grammatically incorrect:

> 1 Invalidated Block Hashes

Should be:

> 1 Invalidated Block Hash
> 2 Invalidated Block Hashes
> 21 Invalidated Block Hash

src/features/blocks/utils/useBlockTable.ts
```diff
@@ -13,53 +13,56 @@ export type TablePagination = {
...
     };
-
-    return res.data;
-  });
+  }
 
   function setTransaction(transaction: TransactionDto[]) {
     selectedTransaction.value = transaction;
+    fetch();
   }
```
0x009922: Instead of triggering re-computation manually (by calling `fetch()`), set up the reactivity in a way so that `fetch` is called automatically whenever a dependency is changed (e.g. using `watch`).

src/pages/BlockDetails.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  it('renders the block-details-card and block-transactions-list correctly', async () => {
+    router.push('/blocks/606600');
+    await router.isReady();
+    const wrapper = mount(BlockDetails, {
+      global: {
+        plugins: [router],
+      },
+    });
+    await wrapper.vm.$nextTick();
+    await new Promise(resolve => setTimeout(resolve, 10));
```
0x009922, line 34: Why this 10ms waiting is needed?

0x009922 (commented):
Glad to see tests!

For now it doesn't seem useful, but the direction is definitely good. Add more tests which assert the correctness or rendering.

src/pages/BlockDetails.spec.ts
horizenight: This is the minimum time that is needed to let the component load. While the `nextTick()` method should await and let the component load for the component but `BlockTable.vue` using a custom load time was the alternative solution.

src/features/blocks/utils/useBlockTable.ts
horizenight: Acknowledged and committed 👍🏼

src/shared/lib/localization/en.json
horizenight: Acknowledged and committed 👍🏼

src/pages/BlockDetails.spec.ts
0x009922: Hmm, can you elaborate on which process within the component needs to load some extra time? Why single `nextTick` is not enough?

0x009922:
Not relevant anymore.

PR #33, opened 2024-07-22
Feature/EW-1 - Add account details page

VladislavPopovSR:
https://app.zenhub.com/workspaces/iroha-v2-60ddb820813b9100181fc060/issues/zh/4


src/pages/Accounts/types.ts
```diff
@@ -0,0 +1,9 @@
+import type { Component } from 'vue';
+
+export interface TableAsset {
+  symbol: string
+  name: string
+  price: number
+  amount: number
+  icon: Component | null
+}
```
0x009922, line 9: You see, Iroha doesn't support `symbol` and `price`. It is a discrepancy between the design and reality. I would ask to stick to reality in the first place, making as many steps towards the design as possible, but to know where to stop. 

You can refer to this schema in the docs: [`AssetDefinition`](https://hyperledger.github.io/iroha-2-docs/reference/data-model-schema.html#assetdefinition). Although it is outdated, it still reflects what actual assets look like in Iroha.
 
And it is still reflected in `dto.ts`, expect the `logo` field. Since in Iroha it is stored as IPFS path, I don't know yet how it's going to work in the explorer. Resolved by the backend? Resolved on the front-end part? Idk, let's leave a stub for now.

TL;DR: remove `TableAsset`, rely only on data from DTOs.

src/pages/Accounts/consts.ts
```diff
@@ -0,0 +1,14 @@
+export const assetsItems = [
+  { label: 'cryptos', value: 'cryptos' },
+  { label: 'nfts', value: 'nfts' },
+];
```
0x009922, line 4: Iroha doesn't have this separation of entities. Let's not impl these tabs for now.

src/features/filter-transactions/TransactionTypeFilter.vue
```diff
@@ -2,32 +2,34 @@
...
-  { label: t('burns'), value: 'burns' },
-  { label: t('grants'), value: 'grants' },
-  { label: t('revokes'), value: 'revokes' },
+  { label: props.defaultOptions ? 'transactions' : 'all', value: props.defaultOptions ? 'transactions' : 'all' },
+  { label: 'transfers', value: 'transfers' },
+  { label: 'mints', value: 'mints' },
+  { label: 'burns', value: 'burns' },
+  { label: 'grants', value: 'grants' },
+  { label: 'revokes', value: 'revokes' },
 ];
```
0x009922, line 33: This depends on a reactive `props.defaultOptions`. Snouldn't it be `computed`?

src/pages/Accounts/AccountDetailsPage.vue
```diff
@@ -0,0 +1,606 @@
...
+import { useErrorHandlers } from '@/shared/ui/composables/useErrorHandlers';
+
+const router = useRouter();
+
+const { handleUnknownError } = useErrorHandlers();
+
+const HASH_BREAKPOINT = 1200;
+const { width } = useWindowSize();
+
+const hashType = computedEager(() => (width.value < HASH_BREAKPOINT ? 'short' : 'medium'));
```
0x009922, line 35: You mentioned yourself that `computedEager` isn't needed in Vue 3.3+

But I see that you haven't updated Vue itself. Can you please bump its version too?

src/pages/Accounts/AccountDetailsPage.vue
```diff
@@ -0,0 +1,606 @@
...
+    handleUnknownError(e);
+  } finally {
+    isFetchingAccount.value = false;
+  }
+});
+
+const assetsFilter = ref('cryptos');
+
+// use account's assets from payload instead or replace with fetching account's assets method
+const assetsTable = useTable(getFakeAssets);
```
0x009922, line 60: There is an already existing mechanism for fake data. It is implemented in a seamless way through the central `http` mock, so that application code isn't tightly coupled to it.

Please, follow the same way here.

src/pages/Accounts/AccountDetailsPage.vue
```diff
@@ -0,0 +1,606 @@
...
+// use account's transactions from payload instead or replace with fetching account's transactions method
+const transactionsTable = useTable(transactionModel.fetchList);
+
+// replace with account's data
+const accountAlias = ref('genesis@genesis.com"');
+const accountSignature = ref('0xF9450D254A66abs3fsf334vds4341cdsas06b30Cfa9c6e7AE1B7598c7172');
+
+const accountTotalValue = ref(875123);
+const accountReservedValue = ref(1234);
+const accountLockedValue = ref(543);
```
0x009922, line 74: Design discrepancy; remove this.

0x009922 (changes requested):
Additionally, to help with async data data in reactive contexts, take a look at our [`@vue-kakuyaku/core`](https://github.com/soramitsu/vue-kakuyaku/tree/master/packages/core) library, produced by yours truly.

src/shared/ui/components/BaseDropdown.vue
```diff
@@ -1,7 +1,6 @@
 <template>
   <div
     class="base-dropdown"
-    role="select"
```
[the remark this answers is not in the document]
VladislavPopovSR: This occurs sonarqube issue. And works the same as before.

src/shared/ui/composables/useErrorHandlers.ts
```diff
@@ -0,0 +1,26 @@
...
+  const notifications = useNotifications();
+
+  function handleUnknownError(error: unknown) {
+    console.error(error);
+
+    const errorMessage = getErrorMessage(error);
+    notifications.error(errorMessage);
+  }
+
+  return { handleUnknownError } as const;
```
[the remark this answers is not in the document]
VladislavPopovSR: It's always better to return const variable. Otherwise the following would be possible 
```
obj.handleUnknownError = null
```

src/features/filter-transactions/TransactionTypeFilter.vue
VladislavPopovSR: This computes only once. I think it's enough

src/shared/ui/components/BaseDropdown.vue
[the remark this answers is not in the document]
0x009922: What issue?

I know that it works the same as before, but probably not from the standpoint of accessibility (which is important to keep).

src/shared/ui/composables/useErrorHandlers.ts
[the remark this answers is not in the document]
0x009922: Are you sure?

```ts
function test() {
    return {
        foo: () => 'foo'
    }
}

const value = test()
value.foo = null  // Type 'null' is not assignable to type '() => string'
```

[Playground](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABFApgZygCgJSIN4BQixiATilCKUoSXYsHHAFyI6IC8AfIgOSNxeREgF8CYghAQZEANwCGAGxApOydFmwEFylADoBasCEWLiAenOIAKgE8ADqt7HTvRDDSIwcKInlo0GABzMHkAI0VVKDhkByd2bkQMUhgwIKEgA)

I just don't see a strong reason to put `as const` for all returns

src/features/filter-transactions/TransactionTypeFilter.vue
0x009922: Then document it! That `defaultOptions` is not supposed to be changed dynamically.

src/shared/ui/components/BaseDropdown.vue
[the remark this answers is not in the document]
VladislavPopovSR: This one:
Снимок экрана 2024-07-25 в 11 19 54

src/shared/ui/composables/useErrorHandlers.ts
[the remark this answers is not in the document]
VladislavPopovSR: My bad. The following would be possible
```
const errorObj = useErrorHandlers()
errorObj.handleUnknownError = () => 1
```

Alright I will remove as

src/shared/ui/components/BaseDropdown.vue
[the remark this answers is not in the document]
0x009922: Ah, okay...

src/pages/Accounts/AccountDetailsPage.vue
```diff
@@ -0,0 +1,460 @@
...
+const router = useRouter();
+
+const { handleUnknownError } = useErrorHandlers();
+
+const HASH_BREAKPOINT = 1200;
+const { width } = useWindowSize();
+
+const hashType = computed(() => (width.value < HASH_BREAKPOINT ? 'short' : 'medium'));
+
+const accountId = computed(() => String(router.currentRoute.value.params['id']));
```
0x009922, line 34: Isn't `params['id']` already a string? If you added `String()` to convert possible `string[]` to a string, it seems to be not a desirable logic.

Consider adding simple assertions using something like [`tiny-invariant`](https://www.npmjs.com/package/tiny-invariant):

```ts
const id = router.currentRoute.value.params['id']
invariant(typeof id === 'string')
return id // TS narrowed the type
```

src/shared/api/dto.ts
0x009922, line 1: This file still has some changes, while it should remain entirely unchanged.

test/vitest.config.ts
0x009922, line 1: You can put this file in the project root, and remove `mergeConfig` - Vitest will extent base Vite config automatically.

test/money-formatter.test.ts
0x009922, line 1: I think placing this file right next to the source file itself (`money-formatters.ts`) would be... easier for review/future maintanence/simpler project organisation? Seems to be no harm.

test/money-formatter.test.ts
```diff
@@ -0,0 +1,9 @@
+import { expect, test } from 'vitest';
+import { numberFormatter } from '../src/shared/utils/money-formatters';
+
+test('format number', () => {
+  expect(numberFormatter(1234567.22312)).toBe('1,234,567.22312');
+  expect(numberFormatter(1234123)).toBe('1,234,123');
+  expect(numberFormatter(0.22312)).toBe('0.22312');
+  expect(numberFormatter(100.1)).toBe('100.1');
+});
```
0x009922, line 9: [`test.each`](https://vitest.dev/api/#test-each)

test/BaseTabs.test.ts
```diff
@@ -0,0 +1,42 @@
+import { test, expect } from 'vitest';
+import { render } from '@testing-library/vue';
```
0x009922, line 2: Why `testing-library`?

I would prefer if you use `@vue/test-utils` directly, because it is a quite good enough official Vue toolkit.

test/BaseTabs.test.ts
```diff
@@ -0,0 +1,42 @@
...
+      },
+      props: {
+        items: defaultTransactionTypeOptions,
+        modelValue: model.value,
+        adaptiveOptions: adaptiveTransactionTypeOptions,
+      },
+    });
+
+    const tabs = document.getElementsByClassName('base-tabs__tab').length;
+    expect(tabs).toBe(applyAdaptiveOptions(width, adaptiveTransactionTypeOptions));
```
0x009922, line 31: This use of `applyAdaptiveOptions` as an expected value some reduces the usefulness of the test. Better to put an exact expected number - test would also become more expressive.

test/BaseTabs.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ref } from 'vue';
+import { adaptiveTransactionTypeOptions } from '../src/pages/Accounts/consts';
+import { applyAdaptiveOptions } from '../src/shared/utils/adaptive-options';
+import { i18n } from '../src/shared/lib/localization';
+
+test('BaseTabs adaptive display correctness', async () => {
+  const model = ref<DefaultTransactionTypeTabs>('transactions');
+
+  // Since rerender method doesn't work as expected due to changing window width
+  // we will render & unmount our component every time
```
0x009922, line 15: Sounds like quite a good reason to run _each_ case in a separate test!

src/widgets/latest-transactions/LatestTransactions.vue
```diff
@@ -22,7 +22,7 @@
         class="latest-transactions__row"
       >
         <TransactionStatus
-          :committed="transaction.committed"
+          :committed="!!transaction.rejection_reason"
```
0x009922, line 25: You see, in the previous form there was a ready-to-use `committed` flag, but now it's the same `!transaction.rejection_reason` in a lot of places. And it's so much easier to make a typo
```suggestion
          :committed="!transaction.rejection_reason"
```

src/shared/config.ts
```diff
@@ -18,3 +19,5 @@ export const langOptions = [
 ];
 
 export const PORTAL_ID = 'header-dropdown-portal';
+
+export const applicationCurrency = ref('$');
```
0x009922, line 23: Preferrably to follow the existing style - define a composable in `ui/composables`

```ts
export function useApplicationCurrency(): Ref<string>
```

src/shared/utils/adaptive-options.ts
0x009922, line 1: These (+ money formatter) utils are clearly ui-related. There is a place for them.

test/BaseTabs.test.ts
0x009922, line 1: You haven't covered that part of `BaseTabs` component which is the most fragile and prone to errors (and therefore is useful to test) - how component reacts to arrows clicks.

src/shared/ui/components/BaseTable.vue
```diff
@@ -91,6 +94,8 @@ interface Props {
   pagination: TablePagination
   items: any[]
   containerClass: string
+  breakpoint?: string | number
+  disabledPagination?: boolean
```
0x009922, line 98: Stronger types, minimal API: 

```suggestion
  pagination?: false | TablePagination
  items: any[]
  containerClass: string
  breakpoint?: string | number
```

src/shared/ui/components/BaseTable.vue
```diff
@@ -183,7 +190,7 @@ const pageSizeModel = computed({
 });
 </script>
 
-<style lang="scss">
+<style lang="scss" scoped>
```
0x009922, line 193: Why?

src/shared/ui/components/BaseHash.vue
```diff
@@ -47,21 +49,29 @@ async function copy() {
...
+      first: props.hash.slice(0, center),
+      second: props.hash.slice(-center + (props.hash.length % 2)),
+    };
   }
+
+  return {
+    t: 'plain',
+    value,
+  };
 });
```
0x009922, line 75: This code doesn't statically ensure that all possible `props.type` values are covered.

You can refactor it either using a `switch` statement, or add a branch with `never` assertion, or use [`ts-pattern`](https://www.npmjs.com/package/ts-pattern) (the most robust option).

0x009922 (changes requested):

src/entities/transaction/model.ts
```diff
@@ -1,23 +1,21 @@
...
+    creation_time: string
+    time_to_live_ms: number | null
+    nonce: null | number
+    metadata: any
   }
+  signatures: Signature[]
+  rejection_reason?: string
 }
 
 function hexToBytes(hex: string) {
```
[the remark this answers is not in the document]
0x009922: Nah, it is due to the commented code below.

src/entities/transaction/model.ts
```diff
@@ -1,23 +1,21 @@
...
-    rejection_reason?: string
+export interface Transaction {
+  committed: boolean
+  block_hash: string
+  block_height: number
+  hash: string
+  payload: {
+    account_id: string
+    instructions: Instruction[]
+    creation_time: string
```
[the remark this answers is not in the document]
VladislavPopovSR: Origin creation_time type on backend is string

src/entities/transaction/model.ts
```diff
@@ -30,28 +28,23 @@ function hexToBytes(hex: string) {
 }
 
 export function mapFromDto(transaction: TransactionDto): Transaction {
-  const instructions =
-    transaction.c.payload.instructions.c?.map((i: string) => Instruction.fromBuffer(hexToBytes(i))) ?? [];
+  const instructions: Instruction[] = [];
+  // TypeError: Invalid argument type in ToBigInt operation (@iroha2_data-model.js:240)
+  //   transaction.c.payload.instructions.c?.map((i: string) => Instruction.fromBuffer(hexToBytes(i))) ?? [];
```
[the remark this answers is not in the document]
VladislavPopovSR: I will mark it for now. Later it will be clear what should I do.

src/features/filter-transactions/model.ts
```diff
@@ -1,2 +1,21 @@
...
+];
+
+export const transactionTypeOptions = [
+  { label: 'all', value: 'all' },
+  { label: 'transfers', value: 'transfers' },
+  { label: 'mints', value: 'mints' },
+  { label: 'burns', value: 'burns' },
+  { label: 'grants', value: 'grants' },
+  { label: 'revokes', value: 'revokes' },
+];
```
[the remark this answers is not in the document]
VladislavPopovSR: lets keep it as is. I will rename it to better understanding. Maybe later we will extend/reduce those filters. They are similar but different entities. Also i18n isn't lost since we use them only in BaseTabs component that has the following label display:
```
{{ $t(item.label) }}
```

src/pages/Accounts/AccountDetailsPage.vue
VladislavPopovSR: Maybe just leave as the following? There won't be cases with array values
```
const id = router.currentRoute.value.params['id']
```

src/entities/transaction/model.ts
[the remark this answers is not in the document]
0x009922: Yes. But you have `mapFromDto` function, which converts the DTO into a structure that is more convenient to use within application. It is a perflect place to convert raw `string` into `Date`.

src/features/filter-transactions/model.ts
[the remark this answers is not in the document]
0x009922: Okay, let's keep it.

> `{{ $t(item.label) }}`

Maybe `i18nKey` instead of `label` would be clearer.

src/pages/Accounts/AccountDetailsPage.vue
0x009922: Assertions generally are helpful in future. Up to you.

src/features/filter-transactions/model.ts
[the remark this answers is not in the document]
0x009922: Okay, let's keep it.

> `{{ $t(item.label) }}`

Maybe `i18nKey` instead of `label` would be clearer.

src/pages/Accounts/AccountDetailsPage.vue
VladislavPopovSR: According to https://stackoverflow.com/a/72839752 seems like we won't have any cases with array values but I will add an assertion anyway👍

src/widgets/latest-transactions/LatestTransactions.vue
VladislavPopovSR: For now we are using TransactionDto type. Later when mapping transaction from dto will be applied we will rely on committed field. Anyway I will correct such expressions.

src/shared/ui/components/BaseHash.vue
VladislavPopovSR: There won't be more cases of displaying so I think it will be enough. In my opinion, adding 1 more package for a better assertion isn't necessary for this case. Anyway, thanks for the docs, this package is new for me :)

src/shared/ui/components/BaseTable.vue
VladislavPopovSR: I think there were troubles with styles during development but now it doesn't affect anything so I will remove it.

src/shared/ui/components/BaseHash.vue
0x009922: Agree that adding a package for only this case is too much.

However, please make static checking exhaustive anyway. Here is an example how to do so with `switch`: [How do I check that a switch block is exhaustive in TypeScript? - Stack Overflow](https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript)

src/entities/transaction/model.ts
```diff
@@ -30,28 +28,25 @@ function hexToBytes(hex: string) {
...
 
-export async function fetchList(params?: PaginationParams): Promise<Paginated<Transaction>> {
-  const res = await http.fetchTransactions(params);
-  const data = res.data.map(mapFromDto);
-
-  return {
-    pagination: res.pagination,
-    data,
-  };
+export async function fetchList(params?: PaginationParams): Promise<Paginated<TransactionDto>> {
```
0x009922, line 50: This function used to return a mapped `Transaction`, not a raw `TransationDto`. Why you changed it? Why it was the way it was, in your opinion?

src/features/filter-transactions/model.ts
```diff
@@ -1,2 +1,23 @@
...
+
+export const defaultTransactionTypeOptions = [
+  { label: i18n.global.t('all'), value: 'all' },
+  { label: i18n.global.t('transfers'), value: 'transfers' },
+  { label: i18n.global.t('mints'), value: 'mints' },
+  { label: i18n.global.t('burns'), value: 'burns' },
+  { label: i18n.global.t('grants'), value: 'grants' },
+  { label: i18n.global.t('revokes'), value: 'revokes' },
+];
```
0x009922, line 23: `t` is reactive, and these values won't update.

src/shared/ui/utils/money-formatters/money-formatters.test.ts
```diff
@@ -0,0 +1,11 @@
+import { expect, test } from 'vitest';
+import { numberFormatter } from './money-formatters';
+
+test.each([
+  [1234567.22312, '1,234,567.22312'],
+  [1234123, '1,234,123'],
+  [0.22312, '0.22312'],
+  [100.1, '100.1'],
+])('format number', (input, expected) => {
```
0x009922, line 9:
```suggestion
])('format %f as %s', (input, expected) => {
```

Better test reports.

test/BaseTabs.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    await nextButton.trigger('click');
+    checkTabs();
+
+    const prevButton = wrapper.find('.base-tabs__arrow');
+
+    await prevButton.trigger('click');
+    checkTabs();
+  }
+
+  wrapper.unmount();
```
0x009922, line 59: Doesn't it unmount automatically?

src/shared/ui/components/BaseTabs.vue
```diff
@@ -1,19 +1,39 @@
 <template>
   <div class="base-tabs">
     <div
-      v-for="(item, i) in props.items"
+      v-if="adaptiveOptions && adaptiveIndexStart !== 0"
+      class="base-tabs__arrow"
```
0x009922, line 5:
```suggestion
      class="base-tabs__arrow"
      data-testid="prev"
```

src/shared/ui/components/BaseTabs.vue
```diff
@@ -1,19 +1,39 @@
...
       class="base-tabs__tab"
       :class="{ 'base-tabs__tab--active': item.value === model }"
       @click="model = item.value"
     >
       {{ item.label }}
     </div>
+
+    <div
+      v-if="adaptiveOptions && adaptiveIndexEnd !== props.items.length"
+      class="base-tabs__arrow"
```
0x009922, line 23:
```suggestion
      class="base-tabs__arrow"
      data-testid="next"
```

test/BaseTabs.test.ts
0x009922: Move this file next to the component itself too, please

test/BaseTabs.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (expectedTabs === 2) {
+    const nextButton = wrapper.find('.base-tabs__arrow');
+
+    await nextButton.trigger('click');
+    checkTabs();
+
+    await nextButton.trigger('click');
+    checkTabs();
+
+    const prevButton = wrapper.find('.base-tabs__arrow');
```
0x009922, line 43: These selectors are confusing - what's the difference between the prev and next?
Use `[data-testid=prev]` instead.

test/BaseTabs.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    await prevButton.trigger('click');
+
+    checkTabs();
+  } else if (expectedTabs !== wrapper.vm.items.length) {
```
0x009922, line 47: Is it possible to rewrite the test without this dynamic comparison with `wrapper.vm.items.length`?

Tests are to be expressive, simple, understood at a glance, without conditional logic, and preferrably without relying too much on the component internals.

vite.config.mts
```diff
@@ -1,10 +1,14 @@
...
 import { defineConfig } from 'vite';
+import path from 'path';
 import vue from '@vitejs/plugin-vue';
 import svg from 'vite-svg-loader';
 
 // https://vitejs.dev/config/
 export default defineConfig({
+  test: {
+    environment: 'jsdom',
+  },
```
0x009922, line 11: It works even without using `import { defineConfig } from 'vitest'`?

src/shared/ui/composables/useApplicationCurrency.ts
```diff
@@ -0,0 +1,22 @@
...
+  USD: '$',
+} as const;
+
+export function useApplicationCurrency() {
+  const applicationCurrency = ref<ApplicationCurrency>('USD');
+
+  function setApplicationCurrency(value: ApplicationCurrency) {
+    // TODO: use local storage
+    applicationCurrency.value = value;
+  }
```
0x009922, line 16: [`useLocalStorage` | VueUse](https://vueuse.org/core/useLocalStorage/#uselocalstorage)

```suggestion
  const applicationCurrency = useLocalStorage<ApplicationCurrency>('app-currency', 'USD');

  function setApplicationCurrency(value: ApplicationCurrency) {
    applicationCurrency.value = value;
  }
```

src/pages/TransactionsList.vue
```diff
@@ -24,12 +26,12 @@
       @set-page="table.setPage($event)"
       @set-size="table.setSize($event)"
     >
-      <template #row="{ item }: { item: Transaction }">
+      <template #row="{ item }: { item: TransactionDto }">
```
0x009922, line 29:
```suggestion
      <template #row="{ item }: { item: Transaction }">
```

vite.config.mts
VladislavPopovSR: As I understood, newer versions of vite support this field. Additionally, it even works without using reference line
```
/// <reference types="vitest" />
import { defineConfig } from 'vite';
```
But I would keep it.

src/shared/ui/composables/useApplicationCurrency.ts
VladislavPopovSR: I also added useApplicationLanguage composable since they are similar.

src/shared/config.ts
```diff
@@ -1,11 +1,9 @@
-import { i18n } from '@/shared/lib/localization';
-
 export const menu = [
-  { label: i18n.global.t('blocks'), to: '/blocks' },
-  { label: i18n.global.t('assets'), to: '/assets' },
-  { label: i18n.global.t('domains'), to: '/domains' },
-  { label: i18n.global.t('accounts'), to: '/accounts' },
-  { label: i18n.global.t('transactions'), to: '/transactions' },
+  { label: 'blocks', to: '/blocks' },
```
0x009922, line 2:
```suggestion
  { i18nKey: 'blocks', to: '/blocks' },
```

src/features/switch-lang/LangDropdown.vue
```diff
@@ -15,26 +15,40 @@
...
+  if (dropdown.isOpen.value) dropdown.toggle();
+
+  setApplicationCurrency(value);
+}
+
+const { locale } = useI18n();
+
+watch(language, () => {
+  locale.value = language.value;
+});
```
0x009922, line 51: Hm, I think this logic is better to move inside of the `useApplicationLanguage` composable.

src/shared/ui/utils/money-formatters/money-formatters.ts
0x009922, line 1: A separate dir `money-formatters` for a single file `money-formatters.ts`. Nesting into a directory is probably redundant.

vite.config.mts
0x009922: Cool! Good to know.

src/shared/ui/composables/useApplicationLanguage.ts
```diff
@@ -0,0 +1,21 @@
+import { useLocalStorage } from '@vueuse/core';
+import { watch } from 'vue';
+import { useI18n } from 'vue-i18n';
+
+export const useApplicationLanguage = () => {
+  const language = useLocalStorage('app-language', 'en');
+  const { locale } = useI18n();
```
0x009922, line 7: I think on first page load these two should be synced too.

src/shared/ui/components/BaseHash.vue
```diff
@@ -47,20 +49,37 @@ async function copy() {
...
+        first: props.hash.slice(0, center),
+        second: props.hash.slice(-center + (props.hash.length % 2)),
+      };
+    }
+    default: {
+      return {
+        t: 'plain',
+        value: props.hash,
+      };
+    }
```
0x009922, line 82: No, you still need to cover the last case, and then, in `default`, add a type guard that `props.type` is `never`.

src/shared/ui/components/BaseHash.vue
```diff
@@ -47,20 +50,40 @@ async function copy() {
...
-    default:
-      return props.hash;
+      return {
+        t: 'two-line',
+        first: props.hash.slice(0, center),
+        second: props.hash.slice(-center + (props.hash.length % 2)),
+      };
+    }
+    default: {
+      return assertUnreachable('Wrong type');
```
0x009922, line 85: No, you still don't get it.

Please re-read the stackoverflow link I've provided before.

src/features/filter-transactions/model.ts
0x009922, line 1: A few problems:

- Unnecessary renamings of exported types
- `label` should be `i18nKey`
- `value` should be linked to the related types, e.g. via `Array<{ i18nKey: string, value: Tab }>`. Now it's just a string, not even `const`

0x009922 (changes requested):

src/shared/ui/utils/assert-unreachable.ts
```diff
@@ -0,0 +1,3 @@
+export function assertUnreachable(x: never): never {
+  throw new Error(`Incorrect BaseHash props.type value : ${x}`);
```
0x009922, line 2: Why this message is put here, in `shared/ui/utils`?

I'd suggest just inlining it in the component itself:
 
```ts
const x: never = props.type
throw new Error(`unexpected props.type: ${String(x)}`)
```

src/features/filter-transactions/model.ts
VladislavPopovSR: Renamed to simpler names✅
By the way, TabItem was the original name. I haven't changed it until now.

0x009922: `TabItem`?

src/features/filter-transactions/model.ts
```diff
@@ -1,2 +1,26 @@
 export type Status = 'committed' | 'rejected' | null;
-export type Tab = 'all' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
+export type DefaultTabs = 'all' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
+export type BlockTabs = 'transactions' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
```
0x009922, line 3:
```suggestion
export type TabDefaultScreen = 'all' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
export type TabBlocksScreen = 'transactions' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
```

src/shared/ui/components/BaseTabs.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    checkTabs();
+
+    await nextButton.trigger('click');
+    checkTabs();
+
+    const prevButton = wrapper.find(`[data-testid="prev"]`);
+    await prevButton.trigger('click');
+
+    checkTabs();
+  } else if (expectedTabs !== 6) {
```
0x009922, line 51: And what if it is 6? Then the test will check nothing?

VladislavPopovSR: According to layouts the max length of tabs is 6. Hence, when expected amount of tabs is 6 we don't display arrows. Otherwise we test arrows swiping correctness.

src/features/filter-transactions/model.ts
VladislavPopovSR: Done

src/shared/ui/components/BaseTabs.test.ts
0x009922: Then it makes sense to ensure that arrows aren't displayed when tabs amount is 6. Otherwise it is confusing.

0x009922 (approved):
CI needs to be fixed

0x009922 (approved):

PR #35, opened 2024-08-12
Feat/Iroha2-5 - Add asset details page

VladislavPopovSR:
https://app.zenhub.com/workspaces/iroha-v2-60ddb820813b9100181fc060/issues/zh/5


src/pages/Assets/AssetDetailsPage.vue
```diff
@@ -0,0 +1,226 @@
...
+    handleUnknownError(e);
+  } finally {
+    isFetchingAsset.value = false;
+  }
+});
+
+const transactionStatus = ref<ftm.Status>(null);
+const transactionType = ref<TabAssetsSort>('recent');
+
+// use asset's transactions from payload instead or replace with fetching asset's transactions method
```
0x009922, line 58: I don't understand this comment. Could you rephrase?

And, should it be a TODO/FIXME?

src/pages/Assets/AssetDetailsPage.vue
```diff
@@ -0,0 +1,226 @@
...
+      class="asset-details__metrics"
+    >
+      <template #default>
+        <div
+          v-if="isFetchingAsset"
+          class="asset-details__metrics_loading"
+        >
+          <BaseLoading />
+        </div>
+        <div v-else>
```
0x009922, line 75:
```suggestion
        <div v-else-if="asset">
```

src/pages/Assets/AssetDetailsPage.vue
```diff
@@ -0,0 +1,226 @@
...
+          <div class="asset-details__metrics-data">
+            <DataField
+              :title="$t('name')"
+              :value="asset?.id.split('#')[0]"
```
0x009922, line 79: Split asset id to a name and domain in a `computed`, please.

src/shared/lib/localization/en.json
0x009922, line 1: Current structure of the localisation file looks messy and not well-organised. Perhaps will need to revisit in future.

src/shared/ui/components/BaseTable.vue
```diff
@@ -115,8 +114,8 @@ const { width } = useWindowSize();
 const PAGINATION_BREAKPOINT = 960;
 
 const items = computed(() => {
-  if (props.loading && props.pagination) {
-    return Array.from({ length: props.pagination.page_size }, () => null);
+  if (props.loading) {
+    return Array.from({ length: props.pagination?.page_size ?? 10 }, () => null);
   }
```
0x009922, line 119: I don't immediately like this assumption of 10 by default. Can we leave it as it was for now?

src/shared/lib/localization/en.json
VladislavPopovSR: I removed unused messages and improved a little bit messages' organization.

src/shared/ui/components/BaseTable.vue
VladislavPopovSR: Fixed✅

src/shared/ui/components/BaseTable.vue
```diff
@@ -115,8 +114,8 @@ const { width } = useWindowSize();
 const PAGINATION_BREAKPOINT = 960;
 
 const items = computed(() => {
-  if (props.loading && props.pagination) {
-    return Array.from({ length: props.pagination.page_size }, () => null);
+  if (props.loading) {
+    return [null];
```
0x009922, line 118: Why this change at all?

src/pages/Accounts/AccountDetailsPage.vue
```diff
@@ -54,13 +54,15 @@ onMounted(async () => {
   }
 });
 
-// use account's assets from payload instead or replace with fetching account's assets method
+// TODO: use assets from account payload instead
+//  or replace with fetching account's assets method
```
0x009922, line 58:
```suggestion
// FIXME: this loads ALL assets, not only belonging to the account.
```

src/pages/Accounts/AccountDetailsPage.vue
```diff
@@ -54,13 +54,15 @@ onMounted(async () => {
...
 const assetsTable = useTable(http.fetchAssetDefinitions);
```
0x009922, line 59: Asset definitions do not belong to accounts anyway. They belong to their domains.

Asset = asset definition + account.

```suggestion
const assetsTable = useTable(http.fetchAssets);
```

src/pages/Accounts/AccountDetailsPage.vue
```diff
@@ -54,13 +54,15 @@ onMounted(async () => {
...
 
 const transactionStatus = ref<ftm.Status>(null);
 const transactionType = ref<ftm.TabBlocksScreen>('transactions');
 
-// use account's transactions from payload instead or replace with fetching account's transactions method
+// TODO: use transactions from account payload instead
+//  or replace with fetching specific transactions method
```
0x009922, line 65:
```suggestion
// FIXME: this loads ALL transactions, not only related to the account
```

src/shared/ui/components/BaseTable.vue
VladislavPopovSR: In order to get back to previous expression. It doesn't change anything since display is the same it was before.

src/pages/Accounts/AccountDetailsPage.vue
VladislavPopovSR: ✅

src/shared/ui/components/BaseTable.vue
0x009922: It used to be that if there is a pagination and no data yet, then it will display `page_size` placeholders. Now it ignores pagination and just displays a single placeholder. So, the display is not the same, or am I missing something?

VladislavPopovSR: We have the following condition 
```
v-if="item && (width >= props.breakpoint || !$slots['mobile-card'])" 
```
so either it one or 10 length nullish array - anyway we will display only one row containing BaseLoading component.

0x009922 (approved):

PR #36, opened 2024-08-15
Feat/Explorer-34 - Add block details page

VladislavPopovSR:
https://app.zenhub.com/workspaces/iroha-v2-60ddb820813b9100181fc060/issues/gh/soramitsu/iroha2-block-explorer-web/34


src/shared/api/index.ts
```diff
@@ -11,6 +11,7 @@ interface HTTPService {
   fetchPeerStatus: () => Promise<Status>
   fetchRoles: () => Promise<Role[]>
   fetchBlocks: (params?: PaginationParams) => Promise<Paginated<BlockShallow>>
+  fetchBlock: (height: number) => Promise<BlockShallow>
```
0x009922, line 14:
```suggestion
  fetchBlock: (heightOrHash: number | string) => Promise<BlockShallow>
```

src/shared/api/http.ts
```diff
@@ -59,6 +61,11 @@ export function fetchBlocks(params?: PaginationParams): Promise<Paginated<BlockS
   return get('/blocks', params);
 }
 
+export async function fetchBlock(height: number): Promise<BlockShallow> {
+  // FIXME: replace with real api
+  return makeBlock();
+}
```
0x009922, line 67: Mocks are used in fake http.

```suggestion
export async function fetchBlock(heightOrHash: number | string): Promise<BlockShallow> {
  return get(`/blocks/${heightOrHash}`)
}
```

src/pages/Blocks/BlockDetailsPage.vue
```diff
@@ -0,0 +1,260 @@
...
+const { width } = useWindowSize();
+
+const transactionHashType = computed(() => (width.value < TRANSACTION_HASH_BREAKPOINT ? 'short' : 'full'));
+
+const blockHeight = computed(() => {
+  const height = router.currentRoute.value.params['height'];
+
+  if (typeof height === 'string') return height;
+
+  return height[0];
```
0x009922, line 32: It should never be an array. Use `invariant` to assert that the parameter should be a single string, the only acceptable value.

src/pages/Blocks/BlockDetailsPage.vue
```diff
@@ -0,0 +1,260 @@
...
+
+const { handleUnknownError } = useErrorHandlers();
+
+const TRANSACTION_HASH_BREAKPOINT = 1200;
+const { width } = useWindowSize();
+
+const transactionHashType = computed(() => (width.value < TRANSACTION_HASH_BREAKPOINT ? 'short' : 'full'));
+
+const blockHeight = computed(() => {
+  const height = router.currentRoute.value.params['height'];
```
0x009922, line 28: It is height or hash.

src/app/router.ts
```diff
@@ -9,8 +9,14 @@ const routes = [
   {
     path: '/blocks',
     name: 'blocks-list',
-    component: () => import('@/pages/BlocksList.vue'),
+    component: () => import('@/pages/Blocks/BlocksList.vue'),
   },
+  {
+    path: '/blocks/:height',
```
0x009922, line 15:
```suggestion
    path: '/blocks/:heightOrHash',
```

0x009922 (commented):
Fine, in general.

BTW I am not sure that creating nested directories in `pages` is needed. Flat structure is always better than nested (Python's Zen!) when there is no strong reason against it.

VladislavPopovSR:
I would keep nested directories since they could contain consts types and so on related to page. Also, is it correct expression for previous block arrow display?
```
isPreviousBlockExists.value = block.value.height > 1;
```

0x009922:
> Also, is it correct expression for previous block arrow display?

Yes, it is.



> I would keep nested directories since they could contain consts types and so on related to the page

Those are to be located in related "features" directory, aren't they?

src/pages/AccountDetailsPage.vue
```diff
@@ -14,6 +14,7 @@ import { useWindowSize } from '@vueuse/core';
 import { format } from '@/shared/lib/time';
 import BaseLoading from '@/shared/ui/components/BaseLoading.vue';
 import { useErrorHandlers } from '@/shared/ui/composables/useErrorHandlers';
+import invariant from 'ts-invariant';
```
0x009922, line 17: `tiny-invariant` is a more standard one.

0x009922 (approved):

PR #39, opened 2024-08-23
Feat/Explorer-38 - Update project using new backend features

VladislavPopovSR:
https://app.zenhub.com/workspaces/iroha-v2-60ddb820813b9100181fc060/issues/gh/soramitsu/iroha2-block-explorer-web/38

BaseTable & table.ts sticky pagination functionality might look messy right now. For now, it works as expected. Later I will implement query params & refactor.


src/shared/api/dto.ts
```diff
@@ -1,183 +1,136 @@
...
-    payload: string
-  }
+import { z } from 'zod';
+
+const paginationSchema = z.object({
+  page: z.number(),
+  per_page: z.number(),
+  // TODO: remove nullable when backend is ready
+  total_pages: z.number().nullable(),
+  total_items: z.number().nullable(),
```
0x009922, line 8: These numbers are all not _always_ numbers. In the schema they are `BigInt`s, which when _too_ large become serialized as strings in JSON.

To handle it properly, we have to use `BigInt` instead :<

It might look something like this:

```ts
const toBigInt = z.number().or(z.string()).pipe(z.coerce.bigint());
```

src/shared/api/dto.ts
```diff
@@ -1,183 +1,136 @@
...
+    })
+    .nullable(),
+  payload: transactionPayloadSchema,
+  signature: z.string(),
+  block_hash: z.string(),
+});
+
+export type TransactionDto = z.infer<typeof transactionSchema>;
+
+const transactionBlockSchema = transactionSchema.omit({ block_hash: true });
```
0x009922, line 103: I would say that transaction without `block_hash` is just a transaction, while the other one is `TransactionWithBlockHash` (as in Scalar).

src/shared/api/dto.ts
```diff
@@ -1,183 +1,136 @@
...
+  .partial();
+
+export type PaginationParams = z.infer<typeof paginationParamsSchema>;
+
+export const accountSchema = z.object({
+  id: z.string(),
+  metadata: z.record(z.string(), z.any()),
+});
+
+export type AccountDto = z.infer<typeof accountSchema>;
```
0x009922, line 32: DTO usually desribe the data that comes from the API directly. If it is transformed to something else on frontend, `Dto` suffix seems extra.

Zod allows combining the definition of DTO with its transformation to a convenient type.

```suggestion
export type Account = z.infer<typeof accountSchema>;
```

src/shared/api/http.ts
```diff
@@ -11,58 +24,46 @@ async function get<T>(path: string, params?: PaginationParams): Promise<T> {
   return res.json();
 }
 
-export function fetchAccounts(params?: PaginationParams): Promise<Paginated<Account>> {
+export function fetchAccounts(params?: PaginationParams): Promise<Paginated<AccountDto>> {
   return get('/accounts', params);
```
0x009922, line 28: Use your schemas to validate and transform data from backend!

```suggestion
  return get('/accounts', params).then((raw) => accountsSchema.parse(raw));
```

src/shared/api/dto.ts
```diff
@@ -1,183 +1,136 @@
...
+
+const transactionPayloadSchema = z.object({
+  authority: z.string(),
+  chain: z.string(),
+  created_at: z.string(),
+  instructions: z.union([
+    z.object({ kind: z.literal('Instructions'), value: z.record(z.string(), z.any()).array() }),
+    z.object({ kind: z.literal('Wasm') }),
+  ]),
+  metadata: z.record(z.string(), z.any()),
```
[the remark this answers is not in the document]
VladislavPopovSR: Using json type occurs an error. I've also tried acircular json schema according to documentation but it didn't help.

Снимок экрана 2024-08-27 в 13 07 12

So I decided to reuse current schema
```
z.record(z.string(), z.any())
```

src/shared/api/dto.ts
VladislavPopovSR: I think we don't need to worry about it for now. The max safe integer is 9007199254740991 what means we are too far away to that number. Currently, backend perfectly accepts type coming from frontend and backwards so lets keep it as is.

0x009922: Agreed, not practical yet.

src/shared/api/dto.ts
```diff
@@ -1,183 +1,154 @@
...
+
+const assetSearchParamsSchema = paginationParamsSchema.extend({
+  owned_by: accountIdSchema.optional(),
+});
+export type AssetSearchParams = z.infer<typeof assetSearchParamsSchema>;
+
+export const assetSchema = z.object({
+  id: assetIdSchema,
+  value: z.union([
+    z.object({ kind: z.literal('Numeric'), value: z.string().transform((value) => BigNumber(value)) }),
```
0x009922, line 50: `BigNumber` is needed be used only here. Remove it from everywhere else.

src/shared/api/dto.ts
```diff
@@ -1,183 +1,154 @@
...
+
+export type Asset = z.infer<typeof assetSchema>;
+
+export const assetDefinitionSchema = z.object({
+  id: assetDefinitionIdSchema,
+  logo: z.string().nullable(),
+  metadata: metadataSchema,
+  mintable: z.enum(['Infinitely', 'Once', 'Not']),
+  owned_by: accountIdSchema,
+  type: z.union([
```
0x009922, line 63: Use `discriminatedUnion` instead.

src/shared/api/dto.ts
```diff
@@ -1,183 +1,154 @@
...
+const transactionSearchParamsSchema = paginationParamsSchema.extend({
+  account: accountIdSchema.optional(),
+});
+export type TransactionSearchParams = z.infer<typeof transactionSearchParamsSchema>;
+
+const transactionPayloadSchema = z.object({
+  authority: z.string(),
+  chain: z.string(),
+  created_at: z.string().transform((x) => new Date(x)),
+  instructions: z.union([
```
0x009922, line 87: Same - use a discriminated union

src/shared/api/dto.ts
```diff
@@ -1,183 +1,154 @@
...
-    /**
-     * Public key's multihash
-     */
-    public_key: string
-    /**
-     * Hex binary
-     */
-    payload: string
-  }
+import { z } from 'zod';
```
0x009922, line 1: Would help to review/refactor later if all items in the module are arranged in some order, e.g. alphabetical.

Currently it's hard to navigate.

src/shared/api/http.ts
```diff
@@ -11,58 +32,69 @@ async function get<T>(path: string, params?: PaginationParams): Promise<T> {
...
-export function fetchAccounts(params?: PaginationParams): Promise<Paginated<Account>> {
-  return get('/accounts', params);
-}
+export async function fetchAccounts(params?: PaginationParams): Promise<Paginated<Account>> {
+  const res = await get<Paginated<Account>>('/accounts', params);
+  accountSchema.array().parse(res.items);
 
-export function fetchAccount(id: string): Promise<Account> {
-  return get(`/accounts/${id}`);
+  return res;
```
0x009922, line 39: 1. No need to assign a type to request response (`get<...>`) if you pass it then to Zod's parsing. It accepts `unknown` and produced a typed result.
2. You parse data with schema, but still return non-validated and, more importantly, untransformed response.
3. Define a reusable schema for `Paginated<T>`

```suggestion
  const res = await get('/accounts', params);
  return paginatedSchema(accountSchema).parse(res)
```

```ts
const paginatedSchema = <T extends z.ZodType>(item: T) =>
    z.object({
        pagination: paginationSchema,
        items: item.array()
    })
```

src/shared/api/dto.ts
```diff
@@ -1,183 +1,154 @@
...
+export const accountSchema = z.object({
+  id: accountIdSchema,
+  metadata: metadataSchema,
+});
+
+export type Account = z.infer<typeof accountSchema>;
+
+const assetSearchParamsSchema = paginationParamsSchema.extend({
+  owned_by: accountIdSchema.optional(),
+});
```
0x009922, line 44: This is a type that doesn't need validation/transformation, therefore, schema looks needless here.

src/shared/api/dto.ts
```diff
@@ -1,183 +1,154 @@
...
+import { z } from 'zod';
+import BigNumber from 'bignumber.js';
+
+const paginationSchema = z.object({
+  page: z.number(),
+  per_page: z.number(),
+  // TODO: remove nullable when backend is ready
+  total_pages: z.number().nullable(),
+  total_items: z.number().nullable(),
+});
```
0x009922, line 10: On front-end side, you can preserve previous format (`page`, `page_size`, `pages`, `total`) by adding `.transform` here, avoiding many extra changes across the front-end app.

vite.config.mts
```diff
@@ -6,6 +6,17 @@ import svg from 'vite-svg-loader';
...
 export default defineConfig({
+  server: {
+    proxy: {
+      '/api-proxy': {
+        target: 'http://127.0.0.1:4000',
+        changeOrigin: true,
+        secure: false,
+        ws: false,
+        rewrite: (path) => '/api/v1' + path.replace(/api-proxy\//, ''),
+      },
```
0x009922, line 17: Seems extra, why not rely on the existing `/api/v1` prefix?


```suggestion
      '/api/v1': 'http://127.0.0.1:4000',
```

src/shared/api/dto.ts
VladislavPopovSR: Alphabetical order won't work here since some entities include other entities. For now it has almost similar order as HTTPService interface. Additionally, I left pagination related and shared types in the beginning of the file.

src/features/assets/index.ts
```diff
@@ -0,0 +1,7 @@
+export function getAssetDomain(assetId: string) {
+  return assetId.split('@')[1];
+}
+
+export function getAssetName(assetId: string) {
+  return assetId.split('#')[0];
+}
```
0x009922, line 7: Actually, it might be useful to introduce classes for some IDs.

Backend gives IDs in a format Iroha supports it:

- `signatory@domain` for accounts
- `name#domain` for asset _definitions_
- `asset#asset_domain#signatory@account_domain` for assets (`asset##signatory@domain` - when both the asset definition and the account are in the same domain)

From frontend side, it is sometimes needed to display it in an original form, but sometimes it is necessary to decompose/compose back.

So, I propose to introduce classes `AccountId`, `AssetDefinitionId`, and `AssetId` and integrate them with Zod.

They might look something like this:

```ts
class AssetDefinitionId {
    public construction(name: string, domain: string) {
        // ...
    }

    public get name(): string {}

    public get domain(): string {}

    public toString(): string {} // back to `name#domain`
}

const assetDefinitionSchema = z.string().transform((ctx) => {
    // parse, through error via ctx
    const name, domain

    return new AssetDefinitionId(name, domain)
})
```

VladislavPopovSR: For now I use only AssetDefinitionId and AssetId. AccountId will be implemented later if needed.

0x009922:
Closed in favor of #41

PR #41, opened 2024-09-03
Feature/Explorer 38 - Update project using new backend features

VladislavPopovSR:
https://app.zenhub.com/workspaces/iroha-v2-60ddb820813b9100181fc060/issues/gh/soramitsu/iroha2-block-explorer-web/38

BaseTable & table.ts sticky pagination functionality might look messy right now. For now, it works as expected. Later I will implement query params & refactor.


src/shared/api/dto.ts
```diff
@@ -1,183 +1,228 @@
...
+import { z } from 'zod';
+import BigNumber from 'bignumber.js';
+import invariant from 'tiny-invariant';
+
+const paginationSchema = z.object({
+  page: z.number(),
+  per_page: z.number(),
+  // TODO: remove nullable when backend is ready
+  total_pages: z.number().nullable(),
+  total_items: z.number().nullable(),
```
0x009922, line 10: Not nullable anymore

src/shared/api/dto.ts
```diff
@@ -1,183 +1,228 @@
...
-  interface Paginated<T> {
-    pagination: {
-      page: number
-      page_size: number
-      total: number
-    }
-    data: T[]
+class AssetDefinitionId {
+  private assetName = '';
+  private assetDomain = '';
```
0x009922, line 42:
```suggestion
  private name: string;
  private domain: DomainId;
```

src/shared/api/dto.ts
```diff
@@ -1,183 +1,228 @@
...
+    return this.assetName + '#' + this.assetDomain;
   }
+}
 
-  type AssetValue =
-    | Tagged<'Quantity', string>
-    | Tagged<'BigQuantity', string>
-    | Tagged<'Fixed', string>
-    | Tagged<'Store', JsonObject>;
+export function transformToAssetDefinitionId(assetDefinitionId: string): AssetDefinitionId {
```
0x009922, line 60: In Zod, it would be idiomatic to use [`superRefine`](https://zod.dev/?id=superrefine), actually.

src/shared/api/dto.ts
```diff
@@ -1,183 +1,228 @@
...
-  }
+  return new AssetDefinitionId(asset, asset_domain);
+}
 
-  type Mintable = 'Once' | 'Infinitely' | 'Not';
+class AssetId {
+  private assetName = '';
+  private assetDomain = '';
+  private accountSignatory = '';
+  private accountDomain = '';
```
0x009922, line 73:
```suggestion
  private account: AccountId;
  private definition: AssetDefinitionId;
```

where both are also classes.

src/shared/api/dto.ts
```diff
@@ -1,183 +1,228 @@
...
-    id: string
-    accounts: Account[]
-    asset_definitions: AssetDefinition[]
-    logo: null | string
-    metadata: any
-    /**
-     * amount of triggers, always 0 for now
-     */
-    triggers: number
+  public get account_domain() {
```
0x009922, line 91: We don't have ESLint being setup properly, I guess, but it is common across the codebase to use `camelCase`.

```suggestion
  public get accountDomain() {
```

src/shared/ui/components/BaseTable.vue
```diff
@@ -86,15 +86,16 @@ import { computed } from 'vue';
...
+import type { Pagination } from '@/shared/api/dto';
 
 interface Props {
   loading: boolean
-  pagination?: TablePagination | null
+  pagination?: Pagination | null
   items: T[]
   containerClass: string
   breakpoint?: string | number
+  sticky?: boolean
```
0x009922, line 98: What _sticky_ means? Can it be a name that describes the behaviour better?

src/shared/ui/components/BaseTable.vue
```diff
@@ -125,9 +126,23 @@ const segmentInfo = computed(() => {
...
+
+    const end = start - props.items.length + 1;
+
+    return `${start}—${end} of ${p.total_items}`;
+  }
+
+  const start = (p.page - 1) * p.per_page + 1;
+  const end = p.page * p.per_page;
+
+  return `${start}—${end > p.total_items ? p.total_items : end} of ${p.total_items}`;
```
0x009922, line 145: This functionality seems a bit complicated. It would help to separate the complicated logic into some external function and cover it with unit tests thoroughly.

src/shared/ui/components/BaseTable.vue
0x009922, line 1: This file in general has some non-trivial logic of direct/reverse pagination which is essential to test, imo. It could additionally help to make the implementation clearer.

But it's good to do in a separate PR, having a smaller scope.

src/shared/ui/composables/useErrorHandlers.ts
```diff
@@ -18,5 +19,11 @@ export const useErrorHandlers = () => {
     notifications.error(errorMessage);
   }
 
-  return { handleUnknownError };
+  function handleZodError(error: ZodError) {
+    console.error(error);
+
+    notifications.error(error.errors[0].code);
+  }
```
0x009922, line 26: It can be a single universal error handler that dynamically checks whether `error: unknown` is e.g. an `instanceof ZodError`.

0x009922: Also, there is `.format` method on `ZodError`, please check the docs.

src/shared/api/dto.ts
```diff
@@ -1,183 +1,228 @@
-/* eslint-disable no-unused-vars */
-/* eslint-disable camelcase */
-import type { JsonObject } from 'type-fest';
+import { z } from 'zod';
+import BigNumber from 'bignumber.js';
+import invariant from 'tiny-invariant';
+
+const paginationSchema = z.object({
```
0x009922, line 5: Tip: in TypeScript, there are _runtime_ and _type_ independent namespaces. So, you can have a runtime item (variable, function) and a type item to have the same name.

Therefore, you can have a `const Pagination` that is a Zod schema, and a `type Pagination` that is `z.infer<typeof Pagination>`.

I would strongly advocate to this rename because having `xxxSchema` for all schemas is kind of verbose.

```suggestion
const Pagination = z.object({
```

src/shared/api/dto.ts
0x009922, line 1: Would  make more sense to rename this module to `schemas.ts`.

src/shared/api/index.ts
```diff
@@ -1,19 +1,30 @@
...
+} from '@/shared/api/dto';
+
 interface HTTPService {
   fetchAccounts: (params?: PaginationParams) => Promise<Paginated<Account>>
   fetchAccount: (id: string) => Promise<Account>
-  fetchAssets: (params?: PaginationParams) => Promise<Paginated<Asset>>
-  fetchAsset: (definition_id: string, account_id: string) => Promise<Asset>
-  fetchAssetDefinitions: (params: PaginationParams) => Promise<Paginated<AssetDefinition>>
+  fetchAssets: (params?: AssetSearchParams) => Promise<Paginated<Asset>>
+  fetchAsset: (id: string) => Promise<Asset>
```
0x009922, line 19: If you have a chance to use stronger types, please do!

```suggestion
  fetchAsset: (id: AssetId) => Promise<Asset>
```

src/shared/api/index.ts
```diff
@@ -1,19 +1,30 @@
...
+  Paginated,
+  PaginationParams,
+  TransactionSearchParams,
+  Transaction,
+  PeerStatus,
+} from '@/shared/api/dto';
+
 interface HTTPService {
   fetchAccounts: (params?: PaginationParams) => Promise<Paginated<Account>>
   fetchAccount: (id: string) => Promise<Account>
```
0x009922, line 17:
```suggestion
  fetchAccount: (id: AccountId) => Promise<Account>
```

src/shared/api/index.ts
```diff
@@ -1,19 +1,30 @@
...
-  fetchAssets: (params?: PaginationParams) => Promise<Paginated<Asset>>
-  fetchAsset: (definition_id: string, account_id: string) => Promise<Asset>
-  fetchAssetDefinitions: (params: PaginationParams) => Promise<Paginated<AssetDefinition>>
+  fetchAssets: (params?: AssetSearchParams) => Promise<Paginated<Asset>>
+  fetchAsset: (id: string) => Promise<Asset>
+  fetchAssetDefinitions: (params?: PaginationParams) => Promise<Paginated<AssetDefinition>>
   fetchAssetDefinition: (id: string) => Promise<AssetDefinition>
```
0x009922, line 21:
```suggestion
  fetchAssetDefinition: (id: AssetDefinitionId) => Promise<AssetDefinition>
```

src/shared/api/index.ts
```diff
@@ -1,19 +1,30 @@
...
+  Block,
+  Domain,
+  Paginated,
+  PaginationParams,
+  TransactionSearchParams,
+  Transaction,
+  PeerStatus,
+} from '@/shared/api/dto';
+
 interface HTTPService {
```
0x009922, line 15: Actually, this whole interface doesn't bring benefit anymore. It was needed to ensure consistency between an actual API and a fake one.

So, it would make sense to re-export the `http.ts` module directly.

0x009922 (changes requested):

src/shared/ui/composables/useErrorHandlers.ts
VladislavPopovSR: I use `.parse` method instead of `.safeParse`. The following gives me an empty array:
```
error.format()._errors
```
I would prefer keep using `.parse` method so lets keep it as is. With that, we receive quite clear message about an error.

src/shared/api/dto.ts
VladislavPopovSR: Well, with such naming we can easy understand either it is a type or a zod object. Additionally, when I have
```
export type Account 
export const Account
```
and write `Account` typescript doesn't know what I want to import - type or schema. Hence, instead of auto import I have to check it's correctness every time.

The following might be useful:
```
export type Account
export const account
```

But I would keep it as is.

src/shared/ui/components/BaseTable.vue
VladislavPopovSR: Agreed. Later I will separate it into table and pagination components. Also it will be easier to write unit tests.

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,308 @@
...
+
+export class DomainId {
+  private domain: string;
+  constructor(domain: string) {
+    this.domain = domain;
+  }
+  public getDomain() {
+    return this.domain;
+  }
+}
```
0x009922, line 43: `DomainId` doesn't have any logic and could be just a (branded) `string`.

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,308 @@
...
+
+export class AccountId {
+  private signatory: string;
+  private accountDomain: string;
```
0x009922, line 47:
```suggestion
  private domain: DomainId;
```

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,308 @@
...
+  }
+  public get assetDefinition() {
+    return this.definition.toString();
+  }
+  public get signatory() {
+    return this.account.getSignatory();
+  }
+  public get accountDomain() {
+    return this.account.getAccountDomain();
+  }
```
0x009922, line 119: Seems to be too verbose for no good reason.

It's fine to make the `account` and `definition` fields `readonly public` and access them directly.

Same applies to other classes.

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,308 @@
...
+  public toString() {
+    return this.definition.toString() + '#' + this.account.toString();
```
0x009922, line 121: The logic is a bit different.

If the definition's domain equals to the account's domain, the ID is shortened. Please refer to the API docs where I gave an example for `AssetId` model.

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,308 @@
...
+  const [asset, asset_domain] = assetDefinitionId.split('#');
+
+  return new AssetDefinitionId(asset, asset_domain);
+}
+
+export class AssetId {
+  private account: AccountId;
+  private definition: AssetDefinitionId;
+
+  public constructor(data: { asset: string, asset_domain: string, signatory: string, account_domain: string }) {
```
0x009922, line 100:
```suggestion
  public constructor(account: AccountId, definition: AssetDefinitionId) {
```

src/shared/api/dto.ts
0x009922: > and write `Account` typescript doesn't know what I want to import - type or schema

TypeScript perfectly knows what you are referring to depending on _where_ you are referring to it - in runtime or in type level.

> Hence, instead of auto import I have to check it's correctness every time.

What's the problem? It just a single `import { Account }` for both cases.

I would insist on this uniform naming.

However, a problem could occur is with classes such as `AccountId`, because classes are both types of the instance and also a value of the constructor. In this case having a `Schema` suffix is fine, imo.

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,308 @@
...
+    account_domain,
+  };
+
+  return new AssetId(data);
+}
+
+export const accountSchema = z.object({
+  id: z
+    .string()
+    .superRefine((val, ctx) => {
```
0x009922, line 142: Sorry for introducing confusion with `superRefine`. You can use `(val, ctx)` in `.transform` too, thus combining validation with parsing.

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,308 @@
...
+  hash: z.string(),
+  height: z.number(),
+  created_at: z.string().transform((x) => new Date(x)),
+  prev_block_hash: z.string().nullable(),
+  transactions_hash: z.string(),
+  transactions_rejected: z.number(),
+  transactions_total: z.number(),
+  consensus_estimation: z.object({
+    ms: z.number().min(0),
+  }),
```
0x009922, line 291: I would suggest making a separate `Duration` schema, as it might change in future and is used in several places.

```suggestion
  consensus_estimation: Duration,
```

Same for `Date`.

0x009922:
I have run frontend locally with the backend and collected some notes. Applicable both to frontend and backend. Leaving it here as a paper trail:

- Shift to `height`, not `hash`, as a primary Block ID, and use it everywhere. Return block hash only for `GET /blocks*` endpoints
- Account ID: shorten everywhere, **except** account details page (now the opposite)
- On each details page, print metadata as a raw JSON
- Blocks/Transactions/Instructions - handle "Age" uniformly. Print it as "12 secs ago" with a specific timestamp on hover (fine to do with HTML `title` attr for now)
- Transactions table: print authority and the kind of `instructions` (let's rename it to "Kind of Executable" for clarity) - is it `Instructions` or `WASM`
- Uniform table for Transactions and Instructions, everywhere:
	- `GET /transactions`: `authority`, `block` (height), `status`
	- `GET /instructions`: `authority`, `block`, `kind`
	- Has two sections in filters: Transactions (all, committed, rejected) and Instructions (all, then by kind)
- By pages
	- Blocks list: "Transactions" - add how many rejected ones, e.g. `4 (1 rejected)`
	- Account details
		- add table with owned domains
	- Assets page
		- currently fetches _assets_, not asset _definitions_; fix and print all relevant information from that endpoint
- Missing:
	- Transaction details page
	- Domain details page
	- TX filters aren't working yet

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,299 @@
...
+      definition.getDomain() ? definition.getDomain() : account.getAccountDomain()
+    );
+  }
+
+  public getAccount() {
+    return this.account;
+  }
+  public getDefintion() {
+    return this.definition;
+  }
```
0x009922, line 111: Please remove getters, make fields public, and access directly

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,299 @@
...
+    if (val.split('@').length !== 2) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.invalid_literal,
+        message: 'Invalid accountId format',
+        expected: 'signatory@account_domain format',
+        received: val,
+      });
+    }
+
+    return transformToAccountId(val);
```
0x009922, line 142: No need for a dedicated `transformToAccountId` function, just introduces unnecessary jump in code. It could be inlined here.

src/shared/api/schemas.ts
0x009922, line 1: Naming between types and schemas is still not the same. Please make it so. (e.g. `pagination` vs `Pagination` to `Pagination` for both)

0x009922 (changes requested):

src/shared/api/schemas.ts
VladislavPopovSR: I also use every such transform function in components. In example, `transformToAccountId` used in `AccountDetailsPage` in order to map route param. So it's better to have a reusable function.

src/shared/api/schemas.ts
VladislavPopovSR: Done✅

src/shared/api/schemas.ts
VladislavPopovSR: This is already resolved✅

src/shared/ui/components/BaseTable.vue
VladislavPopovSR: I have already renamed it to reversed✅

src/shared/api/schemas.ts
0x009922: Reuse the schema itself!

src/pages/AccountDetailsPage.vue
```diff
@@ -1,20 +1,20 @@
...
 import BaseHash from '@/shared/ui/components/BaseHash.vue';
 import TransactionStatus from '@/entities/transaction/TransactionStatus.vue';
-import { transactionModel } from '@/entities/transaction';
 import { useWindowSize } from '@vueuse/core';
 import { format } from '@/shared/lib/time';
 import BaseLoading from '@/shared/ui/components/BaseLoading.vue';
 import { useErrorHandlers } from '@/shared/ui/composables/useErrorHandlers';
-import invariant from 'tiny-invariant';
+import type { Account } from '@/shared/api/schemas';
+import { AccountIdSchema } from '@/shared/api/schemas';
```
0x009922, line 17:
```suggestion
import { type Account, AccountIdSchema } from '@/shared/api/schemas';
```

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,252 @@
...
+  id: AccountIdSchema,
+  metadata: Metadata,
+  owned_assets: z.number(),
+  owned_domains: z.number(),
+});
+
+export type Account = z.infer<typeof Account>;
+
+export interface AssetSearchParams extends PaginationParams {
+  owned_by?: string
```
0x009922, line 162:
```suggestion
  owned_by?: AccountId
```

Stronger types!

Tip: if you also add `toJSON()` method to `AccountId` class with `toString()` inside, it will automatically stringify on serialisation without manual intervention.

src/shared/api/schemas.ts
```diff
@@ -0,0 +1,252 @@
...
+  metadata: Metadata,
+  owned_by: AccountIdSchema,
+  accounts: z.number(),
+  assets: z.number(),
+});
+
+export type Domain = z.infer<typeof Domain>;
+
+export interface TransactionSearchParams extends PaginationParams {
+  authority?: string
```
0x009922, line 199:
```suggestion
  authority?: AccountId
```

src/pages/AccountDetailsPage.vue
```diff
@@ -27,43 +27,48 @@ const hashType = computed(() => (width.value < HASH_BREAKPOINT ? 'short' : 'medi
...
 onMounted(async () => {
   try {
     isFetchingAccount.value = true;
     account.value = await http.fetchAccount(accountId.value);
 
     if (account.value) {
-      await Promise.all([assetsTable.fetch(), transactionsTable.fetch()]);
+      await Promise.all([
+        assetsTable.fetch({ owned_by: accountId.value.toString() }),
+        transactionsTable.fetch({ authority: accountId.value.toString() }),
```
0x009922, line 47:
```suggestion
        assetsTable.fetch({ owned_by: accountId.value }),
        transactionsTable.fetch({ authority: accountId.value }),
```

src/pages/AssetDetailsPage.vue
```diff
@@ -24,29 +25,22 @@ const { width } = useWindowSize();
 
 const hashType = computed(() => (width.value < HASH_BREAKPOINT ? 'medium' : 'full'));
 
-const assetName = computed(() => {
+const assetDefinitionId = computed(() => {
   const name = router.currentRoute.value.params['id'];
+  const rest = router.currentRoute.value.hash;
```
0x009922, line 30: You should find a way to put `#` in the route in some escaped way (urlencode). It should not be a part of `route.hash`.

src/pages/AssetDetailsPage.vue
```diff
@@ -24,29 +25,22 @@ const { width } = useWindowSize();
...
 
-  invariant(typeof name === 'string', 'Expected string');
+  const assetId = AssetIdSchema.parse(name + rest);
```
0x009922, line 32: Why parsing `:id` in the route as `AssetId`? It is supposed to be `AssetDefinitionId`, e.g. `roses#wonderland`.

src/pages/AssetDetailsPage.vue
```diff
@@ -59,13 +53,13 @@ onMounted(async () => {
 const transactionStatus = ref<ftm.Status>(null);
 
 // FIXME: this loads ALL transactions, not only related to the asset
-const transactionsTable = useTable(transactionModel.fetchList);
+const transactionsTable = useTable(http.fetchTransactions);
```
0x009922, line 56: Let's remove this table from the asset page altogether. We cannot implement it in the nearest future.

0x009922 (changes requested):

src/shared/api/schemas.ts
VladislavPopovSR: Thanks, thats useful. By the way, I haven't used until yet approach to pass an object in the url so it's automatically uses toString method and converts our classes into normal strings.

src/pages/AccountDetailsPage.vue
VladislavPopovSR: This is ESLint approach to separate imported types and values. Should I really manually change it or maybe there is new eslint rule?

0x009922: If ESLint does it automatically, than it's fine.

Maybe if we change rules it can patch it automatically though.

src/pages/AssetsList.vue
```diff
@@ -24,14 +24,14 @@
       <template #row="{ item }">
         <div class="assets-list-page__row">
           <BaseLink
-            :to="`/assets/${item.id}`"
+            :to="`/assets/${encodeURIComponent(item.id.definition.toString())}`"
```
0x009922, line 27: As it is a common use case both for `AssetId` and `AssetDefinitionId`, maybe worth adding a method like `toStringEscaped()` to both classes. (though only definition id needs it)

0x009922 (approved):

PR #43, opened 2024-09-17
Feat/Explorer-40 - Add transaction details page

VladislavPopovSR:
https://app.zenhub.com/workspaces/iroha-v2-60ddb820813b9100181fc060/issues/gh/soramitsu/iroha2-block-explorer-web/40

### DIFF
--- a/src/pages/TransactionDetailsPage.vue
+++ b/src/pages/TransactionDetailsPage.vue
@@ -1,0 +1,278 @@
+<script setup lang="ts">
+import { useRouter } from 'vue-router';
+import { computed, ref, watch } from 'vue';
+import * as http from '@/shared/api';
+import BaseContentBlock from '@/shared/ui/components/BaseContentBlock.vue';
+import { useWindowSize } from '@vueuse/core';
+import { useErrorHandlers } from '@/shared/ui/composables/useErrorHandlers';
+import BaseLoading from '@/shared/ui/components/BaseLoading.vue';
+import DataField from '@/shared/ui/components/DataField.vue';
+import invariant from 'tiny-invariant';
+import TimeIcon from '@/shared/ui/icons/clock.svg';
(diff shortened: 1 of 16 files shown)


src/shared/lib/time.ts
```diff
@@ -21,7 +23,11 @@ function formatXX(item: number) {
   return item < 10 ? '0' + item : item;
 }
 
-export function format(timestamp: Date | string) {
+export function getUTCFormat(date: Date) {
+  return format(date, `MMM-dd-yyyy hh:mm:ss a 'UTC'`);
+}
```
0x009922, line 28:
```suggestion
export function formatUTC(date: Date) {
  return format(date, `MMM-dd-yyyy hh:mm:ss a 'UTC'`);
}
```

Please add a simple unit test whenever there is a chance to test something not by looking at the browser page.

src/pages/TransactionDetailsPage.vue
```diff
@@ -0,0 +1,276 @@
...
+
+              <DataField
+                :title="$t('transactions.nonce')"
+                :value="transaction.nonce"
+              />
+            </div>
+            <div class="transaction-details__info-row">
+              <DataField
+                :title="$t('transactions.block')"
+                :hash="String(transaction.block)"
```
0x009922, line 123: Why this prop is called `:hash`? It doesn't make any sense here.

VladislavPopovSR: With `:hash` prop it uses `BaseHash` component otherwise it uses `span`. Probably I can make it a boolean. Should I?

0x009922: Why `<BaseHash>` for `transaction.block`?

0x009922 (approved):

PR #45, opened 2024-09-24
Feat - Add tx filters

VladislavPopovSR:
`TransactionList`, `AccountDetailsPage`, `BlockDetailsPage` now have tx filters by kind and status.

Additionally solved:
- Shorten AccountId;
- Add transactions quantity (including rejected) on BlocksList , e.g. "3 (2 rejected)";
- Add owned domains section on AccountDetailsPage;
- On AssetList fetch assets definitions instead;
- On TransactionsList print authority and the kind of instructions (let's rename it to "Kind of Executable" for clarity) - is it Instructions or WASM.


src/shared/api/schemas.ts
```diff
@@ -224,7 +225,7 @@ export const Transaction = z.object({
   block: z.number(),
   created_at: Timestamp,
   executable: z.enum(['Instructions', 'Wasm']),
-  status: z.enum(['Committed', 'Rejected']),
+  status: TransactionStatus,
 });
```
0x009922, line 229: For `DetailedTransaction`, use [`.extend`](https://zod.dev/?id=extend) from `Transaction`

src/shared/api/schemas.ts
```diff
@@ -269,6 +270,14 @@ export const PeerStatus = z.object({
 
 export type PeerStatus = z.infer<typeof PeerStatus>;
 
+export interface InstructionsSearchParams extends PaginationParams {
+  authority?: string
+  kind?: string
+  transaction_hash?: string
+  transaction_status?: 'Committed' | 'Rejected'
```
0x009922, line 277:
```suggestion
  transaction_status?: TransactionStatus
```

src/shared/api/schemas.ts
```diff
@@ -31,6 +31,7 @@ const PaginationParams = z
 export type PaginationParams = z.infer<typeof PaginationParams>;
 
 const Metadata = z.record(z.string(), z.any());
+const TransactionStatus = z.enum(['Committed', 'Rejected']);
```
0x009922, line 34:
```suggestion
const TransactionStatus = z.enum(['Committed', 'Rejected']);
type TransactionStatus = z.infer<typeof TransactionStatus>
```

It is common type used in other places too.

src/pages/TransactionsList.vue
```diff
@@ -84,26 +128,40 @@ const { width } = useWindowSize();
...
+
+async function fetchTransactions() {
   try {
-    await table.fetch();
+    if (shouldUseTransactions.value) await transactionsTable.fetch(reactiveOmit(listState.value, 'kind'));
+    else
+      await instructionsTable.fetch({
+        ...reactiveOmit(listState.value, 'status'),
+        transaction_status: listState.value.status,
+      });
```
0x009922, line 152: It is not a proper use of `reactiveOmit`. It creates a new `reactive()` object on each function call. Reactive objects should only be created within reactive contexts (e.g. during component setup).

You might want to use a plain `omit` here (might be available at `@vueuse/shared`?)

src/pages/TransactionsList.vue
```diff
@@ -84,26 +128,40 @@ const { width } = useWindowSize();
...
 const { handleUnknownError } = useErrorHandlers();
 
-onMounted(async () => {
+const transactionsTable = useTable(http.fetchTransactions, { reversed: true });
+const instructionsTable = useTable(http.fetchInstructions);
+
+const listState = computed(() => ({
+  status: status.value,
+  kind: transactionType.value,
+}));
```
0x009922, line 139: No need to define this.

src/pages/TransactionsList.vue
```diff
@@ -84,26 +128,40 @@ const { width } = useWindowSize();
...
+
+const transactionType = ref<ftm.TabDefaultScreen>('All');
```
0x009922, line 141:
```suggestion
const transactionTab = ref<ftm.TabDefaultScreen>('All');
```

src/pages/TransactionsList.vue
```diff
@@ -84,26 +128,40 @@ const { width } = useWindowSize();
...
+        ...reactiveOmit(listState.value, 'status'),
+        transaction_status: listState.value.status,
+      });
   } catch (e) {
     handleUnknownError(e);
   }
-});
+}
+
+watch(listState, fetchTransactions, { immediate: true });
```
0x009922, line 158:
```suggestion
watch([status, transactionTab], fetchTransactions, { immediate: true });
```

src/pages/BlockDetailsPage.vue
0x009922, line 1: This page contains a big duplicate from `TransactionDetailsPage`. Consider moving the table with transactions/instructions into a separate component.

src/pages/AccountDetailsPage.vue
0x009922, line 1: This component defines the same table with txs/instructions again! Use a dedicated unified component for it, please.

src/features/filter-transactions/model.ts
```diff
@@ -1,6 +1,6 @@
-export type Status = 'committed' | 'rejected' | null;
-export type TabDefaultScreen = 'all' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
-export type TabBlocksScreen = 'transactions' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
+export type Status = 'Committed' | 'Rejected' | null;
```
0x009922, line 1: Remove this `Status` and use `TransactionStatus` from `schemas`. (or you can alias it, but not redefine)

0x009922 (changes requested):

src/pages/TransactionsList.vue
VladislavPopovSR: I think it's better to keep this variable since later I will implement query parameters and it will be necessary to have such variable. Additionally, I will add table properties.

src/pages/TransactionsList.vue
```diff
@@ -58,12 +61,51 @@
...
+      :loading="instructionsTable.loading.value"
+      :pagination="instructionsTable.pagination"
+      :items="instructionsTable.items.value"
+      container-class="transactions-list-page__container"
+      @next-page="instructionsTable.nextPage()"
+      @prev-page="instructionsTable.prevPage()"
+      @set-page="instructionsTable.setPage($event)"
+      @set-size="instructionsTable.setSize($event)"
+    >
+      <template #row="{ item }">
```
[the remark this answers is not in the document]
VladislavPopovSR: Yeah. It's better to have exact payload types for every single instruction type (I have already left todo on instruction schema). Probably as soon as js sdk will have it - I will use it. At least it will be better to display exact data instead of whole json.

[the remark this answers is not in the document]
VladislavPopovSR: I will add it to our todo list.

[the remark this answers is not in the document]
VladislavPopovSR: Sure. Later we will also display more details especially values.

src/features/filter-transactions/model.ts
```diff
@@ -14,19 +32,37 @@ export const sortOptions: TabItem<TabAssetsSort>[] = [
...
-  { i18nKey: 'transactions.revokes', value: 'revokes' },
+  { i18nKey: 'transactions.all', value: 'All' },
+  { i18nKey: 'transactions.transfers', value: 'Transfer' },
+  { i18nKey: 'transactions.mints', value: 'Mint' },
+  { i18nKey: 'transactions.burns', value: 'Burn' },
+  { i18nKey: 'transactions.grants', value: 'Grant' },
+  { i18nKey: 'transactions.revokes', value: 'Revoke' },
+];
+
+export const instructionOptions: TabItem<TabInstructions>[] = [
```
0x009922, line 52: Preferably to use screaming case for consts.

```suggestion
export const INSTRUCTION_OPTIONS: TabItem<TabInstructions>[] = [
```

src/pages/AccountDetailsPage.vue
```diff
@@ -35,22 +26,17 @@ const account = ref<Account | null>(null);
...
   try {
     isFetchingAccount.value = true;
     account.value = await http.fetchAccount(accountId.value);
 
     if (account.value) {
-      await Promise.all([
-        assetsTable.fetch({ owned_by: accountId.value }),
-        transactionsTable.fetch({ authority: accountId.value }),
-      ]);
+      await Promise.all([assetsTable.fetch({ owned_by: accountId.value.toString() })]);
```
0x009922, line 36:
```suggestion
      await assetsTable.fetch({ owned_by: accountId.value.toString() });
```

src/features/filter-transactions/model.ts
```diff
@@ -1,6 +1,24 @@
-export type Status = 'committed' | 'rejected' | null;
-export type TabDefaultScreen = 'all' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
-export type TabBlocksScreen = 'transactions' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
+import type { TransactionStatus } from '@/shared/api/schemas';
+
+export type Status = TransactionStatus | null;
+export type TabDefaultScreen = 'All' | 'Transfer' | 'Mint' | 'Burn' | 'Grant' | 'Revoke';
+export type TabBlocksScreen = 'Transactions' | 'Transfer' | 'Mint' | 'Burn' | 'Grant' | 'Revoke';
```
0x009922, line 5: I don't like this discrepancy between screens. It introduces unnecessary complications in code, and exists due to bad (_subjectively_) design.

More on that in the main comment of my review.

src/shared/ui/components/TransactionsTable.vue
```diff
@@ -0,0 +1,367 @@
...
+import BaseHash from '@/shared/ui/components/BaseHash.vue';
+import BaseLink from '@/shared/ui/components/BaseLink.vue';
+import { useWindowSize } from '@vueuse/core';
+
+const { handleUnknownError } = useErrorHandlers();
+const props = withDefaults(
+  defineProps<{
+    defaultTabs?: boolean
+    authority?: AccountId | null
+    block?: number
```
0x009922, line 26:
```suggestion
    filterBy: { kind: 'authority', id: AccountId } | { kind: 'block', block: number }
```

Stronger types, clearer usage, less chance to make a mistake

src/shared/ui/components/TransactionsTable.vue
```diff
@@ -0,0 +1,367 @@
...
+
+const transactionsTable = useTable(http.fetchTransactions, { reversed: true });
+const instructionsTable = useTable(http.fetchInstructions);
+
+const listState = computed(() => ({
+  status: transactionStatus.value,
+  authority: props?.authority?.toString(),
+  kind: transactionTab.value,
+  block: props?.block,
+}));
```
0x009922, line 53: Use `reactive` instead.

src/shared/ui/components/TransactionsTable.vue
0x009922, line 1: Seems a bit like a mix of two tasks.

What do you think about splitting it into separate components, one for displaying transactions and another for instructions?

0x009922 (changes requested):
Let's make `<TransactionsTable>` in a way that has universal tabs across all views.

Let's say, it consists of 2 main tabs: Transactions vs Instructions.

Transactions have filters only by status. Instructions have filters by their kind _and_ also by status.

Here is a visual sketch

image

src/shared/ui/components/TransactionsTable.vue
VladislavPopovSR: Computed perfectly computes those parameters. Otherwise, we will have to change listState values manually in order not to lose reactivity. Seems like reactive isn't proper or best solution to this.

src/shared/ui/components/TransactionsTable.vue
VladislavPopovSR: Well, first of all I will rename this component to `TransactionsList` since it also contains filters. Sounds better?
The second of all, those `<BaseTable />` components usages are very similar. I think I could create 2 separated tables and then use the following:
```
<TransactionsTable v-if="..." :table="transactionsTable"/>
<InstructionsTable v-else :table="instructionsTable" />
```

But will it really help? I would like to keep as is.

VladislavPopovSR:
I've also changed base url from:
```
const BASE_URL = window.location.toString() + 'api/v1';
```
To:
```
const BASE_URL = window.location.origin.toString() + '/api/v1';
```

I didn't find out this bug immediately because when we run our application, home page has `'/'` location and `BASE_URL` is fine. But if we reload page being on any different route our variable would be `origin + pathname + /api/v1` instead of
correct variant `origin + /api/v1 + pathname`.

src/shared/ui/components/TransactionsList.vue
```diff
@@ -0,0 +1,370 @@
...
+  <div
+    class="transactions-table"
+    :class="{ 'transactions-table_short': props.filterBy?.kind === 'authority' }"
+  >
+    <div class="transactions-table__filters content-row">
+      <TransactionTypeFilter
+        v-if="showInstructions"
+        v-model="transactionTab"
+        :adaptive-options="defaultAdaptiveOptions"
+      />
```
0x009922, line 102: \> `<TransactionTypeFilter`
but
\> `  v-if="showInstructions"`

Transactions filter when showing instructions? Confusing

src/pages/TransactionDetailsPage.vue
```diff
@@ -55,6 +59,25 @@ watch(
...
 );
+
+const transactionTab = ref<ftm.TabInstructions>('All');
+
+const shouldShowAll = computed(() => transactionTab.value === 'All');
+
+const listState = computed(() => ({
+  kind: transactionTab.value,
+  transaction_hash: transactionHash.value,
+}));
```
0x009922, line 70: Use `reactive`, again.
```suggestion
const listState = reactive({
  kind: transactionTab,
  transaction_hash: transactionHash,
});
```

0x009922 (changes requested):
I like some parts, but overall it is still confusing.

Currently:

```
pages/<AccountDetailsPage>
	"Account Transactions" (good: tabs are implemented well!)
		tab: transactions
			<TransactionsList for-account=...>
		tab: instructions
			<TransactionsList for-account=... show-instructions>
				bad: just 5 available kinds, no "all" filter
pages/<TransactionDetailsPage>
	"Transaction instructions" (bad: if it's "WASM", than the title is wrong)
		good: 15 filters, including "all"
		bad: duplicates instructions table from <TransactionsList>
pages/<BlockDetailsPage>
	<TransactionsList for-block=...>
```

Would be less mess:

```
pages/<AccountsShow>
	"Transactions"
		tab: transactions
			<TableTransactions for-account=...>
		tab: instructions
			<TableInstructions for-account=...>
pages/<TransactionsShow>
	"Executable"
		instructions?
			<TableInstructions for-transaction=...>
pages/<BlocksShow>
	<TableTransactions for-block=...>

<TableTransactions>
	for-account | for-block
	inner filters: status
<TableInstructions>
	for-account | for-transaction
	inner filters: status, kind ("all" + 14 specific)
```

src/app/router.ts
```diff
@@ -9,7 +9,7 @@ const routes = [
   {
     path: '/blocks',
     name: 'blocks-list',
-    component: () => import('@/pages/BlocksList.vue'),
+    component: () => import('@/pages/BlocksListPage.vue'),
```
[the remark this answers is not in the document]
VladislavPopovSR: Alright. I also renamed `TransactionsListPage`to `PageTransactionsList` since we also have `TransactionsList` shared component.

src/shared/api/schemas.ts
```diff
@@ -224,16 +226,16 @@ export const Transaction = z.object({
...
   authority: z.string(),
   hash: z.string(),
   block: z.number(),
   created_at: Timestamp,
   executable: z.enum(['Instructions', 'Wasm']),
-  status: z.enum(['Committed', 'Rejected']),
+  status: TransactionStatus,
   rejection_reason: z.record(z.string(), z.any()).nullable(),
   metadata: Metadata,
   nonce: z.number().nullable(),
```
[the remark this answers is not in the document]
VladislavPopovSR: My bad!

VladislavPopovSR:
Finally, I decided add `TransactionsTable` and `InstructionsTable` components. `TransactionsList` is reusable component. It contains filters + query(soon) functionality and displays tables. When on `AccountDetails` page, `InstructionsTable` and switchers between transactions and instructions are shown, otherwise hidden.  Seems not messy :)

src/shared/ui/components/TransactionsList.vue
0x009922, line 1: Please remove this component. It tries to sit on two chairs and looks still clunky.

Having table components for instructions and transactions is enough.

In order to re-use fetching logic of transactions/instructions on pages (block, transaction, account) create reactive composables instead (in needed).

src/pages/TransactionDetails.vue
```diff
@@ -141,31 +161,28 @@ watch(
       </template>
     </BaseContentBlock>
     <BaseContentBlock
-      v-if="transaction"
       class="transaction-details__transactions"
+      :title="$t('transactions.executable')"
```
0x009922, line 165: Could you make it variable? e.g. depending on the executable, make it either "Instructions" or "WebAssembly Smartcontract"

src/pages/PageTransactionsList.vue
0x009922, line 1: Please, make it `TransactionsList` back. Structure of `pages` directory should be consistent.

0x009922 (changes requested):
LGTM overall. Looks like it's almost what we need to have finally. But I am not satisfied with the code organisation.

Apart from comments below, it is confusing to have `TransactionsList` and `TransactionsTable` components. Like, what's the difference? Adds a layer of cognitive overhead.

Please read the Vue styleguide thoroughly. There are comments on component naming and not only.

src/features/filter-transactions/model.ts
```diff
@@ -1,32 +1,48 @@
-export type Status = 'committed' | 'rejected' | null;
-export type TabDefaultScreen = 'all' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
-export type TabBlocksScreen = 'transactions' | 'transfers' | 'mints' | 'burns' | 'grants' | 'revokes';
-export type TabAssetsSort = 'recent' | 'value';
+import type { TransactionStatus } from '@/shared/api/schemas';
+
+export type Status = TransactionStatus | null;
+export type TabAccountTransactions = 'transactions' | 'instructions';
+export type TabInstructions =
+  | ''
```
0x009922, line 6: `''`?..

Makes sense move this into `api/schemas.ts`. It will be there later, and it belongs there. And no non-sensual blind `''`, please.

```suggestion
export type InstructionKind =
```

src/shared/ui/components/TransactionsList.vue
```diff
@@ -0,0 +1,154 @@
...
+);
+
+const transactionsTable = useTable(http.fetchTransactions, { reversed: true });
+const instructionsTable = useTable(http.fetchInstructions);
+
+const listState = reactive({
+  status: null,
+  authority: props.filterBy?.kind === 'authority' ? props.filterBy.id.toString() : '',
+  kind: '' as ftm.TabInstructions,
+  block: props.filterBy?.kind === 'block' ? props.filterBy.block : '',
```
0x009922, line 34: Improper use of `reactive`: it ignores changes to `filterBy`.

I anticipate you saying: "Yes, it is okay. `filterBy` is not changing in the application anyway, it is only set on the first component load"

True. Please make it properly anyway.

Also, don't set empty string for absent values. Better to omit the fields entirely.

src/shared/ui/components/TransactionsTable.vue
```diff
@@ -0,0 +1,216 @@
...
+const props = withDefaults(
+  defineProps<{
+    table: ReturnType<typeof useTable<Transaction>>
+    filterBy?: { kind: 'authority', id: AccountId } | { kind: 'block', block: number } | null
+    hashType: 'short' | 'full'
+  }>(),
+  { filterBy: null }
+);
+
+const table = computed(() => props.table);
```
0x009922, line 20: Redundant - already accessible in the template.

```suggestion
```

src/shared/ui/components/TransactionsTable.vue
```diff
@@ -0,0 +1,216 @@
...
+import TransactionStatus from '@/entities/transaction/TransactionStatus.vue';
+import BaseLink from '@/shared/ui/components/BaseLink.vue';
+import type { useTable } from '@/shared/lib/table';
+import type { AccountId, Transaction } from '@/shared/api/schemas';
+import { computed } from 'vue';
+
+const props = withDefaults(
+  defineProps<{
+    table: ReturnType<typeof useTable<Transaction>>
+    filterBy?: { kind: 'authority', id: AccountId } | { kind: 'block', block: number } | null
```
0x009922, line 14: Replace with bool props that has direct meaning in this component:

- `show-block`
- `show-authority`

src/shared/ui/components/InstructionsTable.vue
```diff
@@ -0,0 +1,261 @@
...
+
+          <div
+            v-if="!props.accounts"
+            class="instructions-table__column-value"
+          >
+            <div class="instructions-table__label">
+              {{ $t('value') }}
+            </div>
+
+            <span class="row-text">{{ Object.entries(item.payload)[0][1] }}</span>
```
0x009922, line 101: What does this magic mean? What does `accounts` prop mean? Impossible to guess if not looking at the invocation.

(and I don't see it even after looking at `<TransactionsList>`)

What could be helpful:

- clearer naming of props, for "dummies" (like me!)
- make a computed for this magic payload manipulation with some name, like `assuminglyAccountIdFromPayload` or idk

src/shared/ui/components/InstructionsTable.vue
```diff
@@ -0,0 +1,261 @@
...
+          <div class="instructions-table__columns">
+            <div
+              v-if="props.allTypes"
+              class="instructions-table__column"
+            >
+              <div class="instructions-table__label">
+                {{ $t('kind') }}
+              </div>
+
+              <span class="row-text">{{ item.kind }}</span>
```
0x009922, line 71: `allTypes`? I don't understand. Maybe..

```suggestion
              v-if="props.showKind"
              class="instructions-table__column"
            >
              <div class="instructions-table__label">
                {{ $t('kind') }}
              </div>

              <span class="row-text">{{ item.kind }}</span>
```

src/shared/ui/components/InstructionsTable.vue
```diff
@@ -0,0 +1,261 @@
...
+import { computed } from 'vue';
+
+const props = defineProps<{
+  table: ReturnType<typeof useTable<Instruction>>
+  accounts?: boolean
+  allTypes?: boolean
+  hashType: 'short' | 'full'
+}>();
+
+const table = computed(() => props.table);
```
0x009922, line 18: Why?

0x009922 (commented):
A definitely like how new table components are working now - just render, no logic, one task, simple, great.

src/pages/AccountDetails.vue
```diff
@@ -0,0 +1,563 @@
...
+
+    if (account.value) {
+      const ownedEntities = [];
+
+      if (account.value.owned_assets) ownedEntities.push(assetsTable.fetch({ owned_by: accountId.value.toString() }));
+      if (account.value.owned_domains) ownedEntities.push(domainsTable.fetch({ owned_by: accountId.value.toString() }));
+
+      if (!ownedEntities.length) return;
+
+      await Promise.all(ownedEntities);
```
0x009922, line 55: Looks like premature optimisation (which is the root of all evil in development) - no need to care about this at this stage of development. Just load everything. (in future, no need to edit this now)

src/pages/AccountDetails.vue
```diff
@@ -0,0 +1,563 @@
...
+
+const transactionsTab = ref<TabAccountTransactions>('transactions');
+
+const shouldShowInstructions = computed(() => transactionsTab.value === 'instructions');
+
+const transactionsTable = useTable(http.fetchTransactions, { reversed: true });
+const instructionsTable = useTable(http.fetchInstructions);
+
+const listState = reactive({
+  status: null,
```
0x009922, line 75: `status` is also a filter within tables, and is separate for each table. It's weird that it is controlled externally.

src/pages/AccountDetails.vue
```diff
@@ -0,0 +1,563 @@
...
+                v-model="listState.kind"
+                :adaptive-options="accountInstructionsAdaptiveOptions"
+              />
+              <TransactionStatusFilter v-model="listState.status" />
+            </div>
+
+            <TransactionsTable
+              v-if="!shouldShowInstructions"
+              :table="transactionsTable"
+              :filter-by="{ kind: 'authority', id: accountId }"
```
0x009922, line 339: There is no `filter-by` prop anymore.

src/pages/AccountDetails.vue
```diff
@@ -0,0 +1,563 @@
...
+
+const shouldShowInstructions = computed(() => transactionsTab.value === 'instructions');
+
+const transactionsTable = useTable(http.fetchTransactions, { reversed: true });
+const instructionsTable = useTable(http.fetchInstructions);
+
+const listState = reactive({
+  status: null,
+  authority: accountId.value.toString(),
+  kind: 'All' as ftm.TabInstructions,
```
0x009922, line 77: It is weird that `kind` filter in `<InstructionsTable>` is controlled by an external component.

src/pages/AccountDetails.vue
0x009922, line 1: Instead of fetching transactions and instructions in this component, it seems better if the relevant components take care of it:

- `TransactionsTable` loads transactions, can filter by account/block, contains status filter within
- `InstructionsTable` is the same, can filter by account/tx, contains its own filters within

0x009922 (changes requested):

0x009922:
Overall this PR is done, and my remarks are related to code organisation.

vite.config.mts
```diff
@@ -6,6 +6,9 @@ import svg from 'vite-svg-loader';
 
 // https://vitejs.dev/config/
 export default defineConfig({
+  define:{
+    __VUE_PROD_HYDRATION_MISMATCH_DETAILS__: false
+  },
```
0x009922, line 11: Why?

0x009922 (approved):

vite.config.mts
VladislavPopovSR: Well updating to Vue 3.4 causes the following warning. Maybe I need to update Vite version as well, I will research that.
Снимок экрана 2024-10-09 в 11 05 58

0x009922 (approved):

PR #47, opened 2024-10-10
Feature - UI changes

VladislavPopovSR:
The following were done:
- Fix dropdown closing;
- For all tables, display rows placeholders to avoid UI shifts on filters/page changes while new data is not loaded yet;
- For all tables, add "no data" filler;
- Unified display of JSON values: monospace, formatted, colored, maybe even interactive;
- Proper "Age" display:
    Adjust precision dynamically (e.g. not "1343 mins 35 secs ago", but "22 hours 23 minutes ago").
    Add a timer to update rendered age according to precision (each N seconds, or minutes, or hours).
    Add a tooltip to display an actual fixed timestamp instead of the age.
- Research what data should show home page's disclaimer. Maybe fetch something.


src/shared/lib/time.ts
```diff
@@ -1,24 +1,23 @@
 import { format } from 'date-fns/format';
+import { useI18n } from 'vue-i18n';
+
+export function getTimeAgo(now: number, dateString: string | Date) {
+  const { t } = useI18n();
```
0x009922, line 5: `useI18n` is a composable that is meant to be called _once_ during `setup`.

So, this properly should be either 

- `useTimeAgo(date: RefOrGetter<Date>): Ref<string>`, or
- `getTimeAgo(now: number, date: Date, t: ...): string` 

And I think it's better to be a composable which "cleverly" tracks current time and updates the value reactively with a reactive timer, all inside.

0x009922 (commented):
Haven't yet tested locally.

VladislavPopovSR:
I updated eslint & typescript dependencies since we are using Vue 3.4 now. In fact, Vue 3.4 supports [shorthand syntax](https://blog.vuejs.org/posts/vue-3-4#v-bind-same-name-shorthand) so I've also added eslint rule in order to highlight an error:

Снимок экрана 2024-10-15 в 14 55 56

Now `format:fix` command will solve such errors.

Also I had to unable `vue/attribute-hyphenation` rule in eslint config. According to the [following issue](https://github.com/vuejs/language-tools/issues/4324) it requires `vue/language-server` newer version. Someone (me too) who doesn't have it's upgraded version will face this error where IDE thinks that we pass two different props :`hash` and `type`.

Снимок экрана 2024-10-15 в 17 53 33

0x009922:
> I updated eslint & typescript dependencies since we are using Vue 3.4 now. In fact, Vue 3.4 supports [shorthand syntax](https://blog.vuejs.org/posts/vue-3-4#v-bind-same-name-shorthand) so I've also added eslint rule in order to highlight an error:

Great!

> In order to display total assets amount should I fetch assets or asset defnitions?

That is about actual assets, not their definitions (I think).

> Also I had to unable `vue/attribute-hyphenation` rule in eslint config. According to the [following issue](https://github.com/vuejs/language-tools/issues/4324) it requires `vue/language-server` newer version. Someone (me too) who doesn't have it's upgraded version will face this error where IDE thinks that we pass two different props :`hash` and `type`.

Hm, okay.

.eslintrc.js
```diff
@@ -24,6 +24,8 @@ module.exports = {
     '@typescript-eslint/explicit-member-accessibility': ['off'],
     '@typescript-eslint/no-parameter-properties': ['off'],
     '@typescript-eslint/unified-signatures': ['off'],
+    'vue/v-bind-style': ['error', 'shorthand', { 'sameNameShorthand': 'always' }],
+    "vue/attribute-hyphenation": ['off'],
```
0x009922, line 28: Add a comment here why you disabled it.

src/shared/ui/composables/useTimeAgo.ts
0x009922, line 1: That doesn't looks idiomatic to the composition api. Look at how composables in VueUse are designed. A more idiomatic version would return a reactive computed value with the desired result.

```ts
function useTimeAgo(date: MaybeRefOrGetter<Date>): Ref<TimeAgo> {}
```

I like the removal of `t` from here. Though, since it is a composable, it is perfectly fine to use `useI18N` next to `useNow`. But if you go the other way, it's better to get rid of notion of `i18nKey` here altogether and replace it with the actual meaningful result value of this composable, e.g.

```ts
type TimeAgo =
  | { precision: 'seconds', seconds: number }
  | { precision: 'minutes', minutes: number }
  // hours, days, months, years
```

I would also suggest using a single value, not a double one. It's simpler. Look at etherscan.io:

image

And finally, it would be an overkill to implement (really a premature optimisation), but it would be an extra challenge! See: if the precision is in minutes or hours, it doesn't make sense to update `now` each second. The update interval can vary depending on the distance from `now`. Hint: [`useIntervalFn`](https://vueuse.org/shared/useIntervalFn/#useintervalfn) supports reactive interval. 

As for testing, I understand that testing a composable with such an API is not super trivial, so I think it's fine to test its core part: the computation of `TimeAgo` based on the distance between a given date and current time. So, you can expose a separate small function and test it trivially:

```ts
declare function timeAgo(target: Date, now: Date): TimeAgo
```

---

upd: I see that having a composable not convenient in tables. Instead, it's better to define a separate component and handle i18n and timers there:

```html
<base-time-ago :value />
```

src/widgets/latest-blocks/LatestBlocks.vue
```diff
@@ -22,9 +25,8 @@
 
             <div class="latest-blocks__time">
               <TimeIcon class="latest-blocks__time-icon" />
-              {{ $t('time.min', [elapsed.allMinutes(block.created_at)]) }}
-              {{ $t('time.sec', [elapsed.seconds(block.created_at)]) }}
-              {{ $t('time.ago') }}
+              <TimeStamp v-bind="getTimeAgo(block.created_at)" />
```
0x009922, line 28:
```suggestion
              <TimeAgo :value="block.created_at" />
```

That would be idiomatic.

.eslintrc.js
0x009922: Actually, as I understand it, it causes only an IDE error, while ESLint check in CLI handles it properly. I would say that a slight inconvenience in IDE has less priority and it more local in scope. So, please revert and use `:hash-type` in actual code.

src/shared/ui/components/BaseInnerBlock.vue
```diff
@@ -55,6 +55,9 @@ function toggle() {
   &__header {
     display: flex;
     align-items: center;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
```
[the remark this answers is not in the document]
VladislavPopovSR: I decided to install and configure `autoprefix` package. It will cover other styles too.

src/shared/lib/time.test.ts
0x009922, line 1: Good. Preferably to make the test even more _unit_ and test the computation function directly. Though these integration tests (component + function + i18n all-in-one) is also useful.

src/shared/ui/composables/useTimeAgo.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  watch(
+    now,
+    () => {
+      const { precision, value } = countTimeDifference(now.value, date);
+
+      if (precision === 'minutes') interval.value = 1000 * 60;
+
+      if (precision !== result.precision) result.precision = precision;
+      if (value !== result.value) result.value = value;
```
0x009922, line 24: both `value` and `precision` are primitive values, therefore Vue _should_ not trigger reactive effects on them being overwritten with the same value. Hence, these `if`s are redundant.

src/widgets/latest-blocks/LatestBlocks.vue
```diff
@@ -22,9 +25,8 @@
 
             <div class="latest-blocks__time">
               <TimeIcon class="latest-blocks__time-icon" />
-              {{ $t('time.min', [elapsed.allMinutes(block.created_at)]) }}
-              {{ $t('time.sec', [elapsed.seconds(block.created_at)]) }}
-              {{ $t('time.ago') }}
+              <TimeAgo :value="block.created_at" />
+              <Tooltip :message="defaultFormat(block.created_at)" />
```
0x009922, line 29: Looks strange - tooltip should be attached to the "time ago" value, shouldn't it?

```html
<tooltip>
  <time-ago :value />
  <template #message> ... </template>
</tooltip>
```

Btw, seems like reasonable to be wrapped in a component which you had removed, i.e. `TimeStamp` (both tooltip and time ago inside)

0x009922 (dismissed):

src/shared/lib/time.test.ts
VladislavPopovSR: Will add later 👍

src/shared/ui/components/TimeStamp.vue
```diff
@@ -0,0 +1,49 @@
...
+import { computed } from 'vue';
+import { useI18n } from 'vue-i18n';
+
+const props = defineProps<{
+  value: Date
+  inverted?: boolean
+}>();
+
+const { t } = useI18n();
+const res = useTimeAgo(props.value);
```
0x009922, line 14: Don't lose reactivity.

```suggestion
const res = useTimeAgo(toRef(props, 'value'));
```

src/shared/ui/composables/useTimeAgo.ts
```diff
@@ -0,0 +1,35 @@
...
+import { reactive, ref, watch } from 'vue';
+import { useIntervalFn } from '@vueuse/shared';
+
+export type TimeAgo =
+  | { precision: 'seconds', value: number }
+  | { precision: 'minutes', value: number }
+  | { precision: 'hours', value: number }
+  | { precision: 'days', value: number };
+
+export function useTimeAgo(date: Date) {
```
0x009922, line 11:
```suggestion
export function useTimeAgo(date: MaybeRefOrGetter<Date>) {
```

src/shared/ui/icons/iroha-logo.png
0x009922, line 1: Please use `src/assets` path

0x009922 (changes requested):

0x009922 (approved):

0x009922 (approved):

Issue #61, opened 2024-10-29
Add nodes amount


VladislavPopovSR:
Add total nodes display on the `HomePage`

Issue #54, opened 2024-10-29
Asset Instances rename 


Mingela:
Rename `Asset Instances` to `Asset Holders`

Issue #49, opened 2024-10-29
Block details changes


Mingela:
- Transactions_hash (should be represented as a state proof or smth)
- Validators signatures (should be stored in the block, not the header); optionally to provide sufficient context for verifying the signatures we may keep track of relevant validators in the explorer backend
- Explicit timezone for a timestamp (both local and UTC on hover or smth)

PR #63, opened 2024-10-30
Feat - Block details changes

VladislavPopovSR:
Solves the following tasks from (#49):  
- Transactions_hash (should be represented as a state proof or smth)
- Explicit timezone for a timestamp (both local and UTC on hover or smth)

Within this pull request I've also added nodes amount (#61) and renamed assets holders (#54)

### DIFF
--- a/src/shared/lib/time.test.ts
+++ b/src/shared/lib/time.test.ts
@@ -4,8 +4,6 @@ import { mount } from '@vue/test-utils';
 import { i18n } from '@/shared/lib/localization';
 import TimeStamp from '@/shared/ui/components/TimeStamp.vue';
 
-// Those tests are written for GMT+3 timezone.
-// In other timezones they won't pass
 test.each([
   [new Date('2024-09-11T07:22:47.157Z'), '11.09.2024 10:22:47 AM GMT+3'],
   [new Date(1730233940352), '29.10.2024 11:32:20 PM GMT+3'],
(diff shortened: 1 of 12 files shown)


src/shared/lib/time.test.ts
```diff
@@ -1,14 +1,23 @@
 import { expect, test } from 'vitest';
-import { formatUTC } from '@/shared/lib/time';
+import { getLocalTime, getUTCTime } from '@/shared/lib/time';
 import { mount } from '@vue/test-utils';
 import { i18n } from '@/shared/lib/localization';
 import TimeStamp from '@/shared/ui/components/TimeStamp.vue';
 
+// Those tests are written for GMT+3 timezone.
+// In other timezones they won't pass
```
0x009922, line 8: Set some fixed timezone for tests according to https://github.com/vitest-dev/vitest/issues/1575#issuecomment-1171085461

0x009922 (changes requested):

src/shared/lib/time.test.ts
VladislavPopovSR: Great, I didn't know about it before! I decided to do the following https://github.com/vitest-dev/vitest/issues/1575#issuecomment-1439286286

test-globals.ts
```diff
@@ -0,0 +1,3 @@
+export const setup = () => {
+  process.env.TZ = 'Europe/Moscow';
```
0x009922, line 2: Please set to some other timezone so that we can see it being reflected in tests for real.

0x009922 (changes requested):

0x009922 (approved):

PR #64, opened 2024-11-13
Refactor - Update fetching

VladislavPopovSR:
- Integrated `vue-kayuaky/core` package;
- Created separated `BasePagination` component instead;
- Corrected styles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "@iroha2/data-model": "^2.0.2",
     "@ngneat/falso": "^6.1.0",
     "@soramitsu-ui/icons": "^0.1.0",
+    "@vue-kakuyaku/core": "^0.4.3",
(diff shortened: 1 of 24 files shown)


VladislavPopovSR:
Now we have the following fetching approach:
1. Create reactive `listState` - our query params
1.1 Create fetching function e.g. if some params should be ommitted
1.2 Write watch where resetting page if necessary (e.g. when we not on the initial page and apply filters)
2. Create scope. Some key's might be conditional
3. Write and use computed variables to access scope data

VladislavPopovSR:
I've also downgraded pnpm version 9.12.1 -> 8.11.0 (back to before) since jenkins was giving the following issue and people advise downgrading pnpm as one of the solution to this. I guess we can keep it for now
[
Снимок экрана 2024-11-14 в 01 15 01
](url)

PR #67, opened 2024-11-14
Feat - Add vue-kakuyaku

VladislavPopovSR:
- Integrated `vue-kayuaky/core` package;
- Created separated `BasePagination` component instead;
- Corrected styles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "@iroha2/data-model": "^2.0.2",
     "@ngneat/falso": "^6.1.0",
     "@soramitsu-ui/icons": "^0.1.0",
+    "@vue-kakuyaku/core": "^0.4.3",
     "@vueuse/core": "^9.1.1",
     "@vueuse/shared": "10.1.2",
     "bignumber.js": "^9.1.2",
--- a/package.json
+++ b/package.json
@@ -33,8 +33,9 @@
     "zod": "^3.23.8"
   },
   "devDependencies": {
+    "@stylistic/eslint-plugin-ts": "^2.11.0",
     "@types/node": "^18.7.14",
-    "@typescript-eslint/eslint-plugin": "^7.18.0",
+    "@typescript-eslint/eslint-plugin": "8.15.0",
     "@typescript-eslint/parser": "^7.18.0",
     "@vitejs/plugin-vue": "^5.0.0",
     "@vue/test-utils": "^2.4.6",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@stylistic/eslint-plugin-ts": "^2.11.0",
     "@types/node": "^18.7.14",
     "@typescript-eslint/eslint-plugin": "8.15.0",
-    "@typescript-eslint/parser": "^7.18.0",
+    "@typescript-eslint/parser": "8.15.0",
     "@vitejs/plugin-vue": "^5.0.0",
     "@vue/test-utils": "^2.4.6",
     "autoprefixer": "^10.4.20",
--- a/package.json
+++ b/package.json
@@ -49,11 +49,11 @@
     "prettier-eslint": "^16.3.0",
     "prettier-eslint-cli": "^8.0.1",
     "sass": "^1.54.6",
-    "typescript": "^5.0.4",
+    "typescript": "5.6.3",
     "vite": "5.2.6",
     "vite-svg-loader": "^3.4.0",
     "vitest": "^2.0.4",
     "vue-eslint-parser": "^9.4.3",
-    "vue-tsc": "^2.0.6"
+    "vue-tsc": "^2.1.10"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "iroha2-block-explorer-web",
-  "packageManager": "pnpm@9.12.1",
+  "packageManager": "pnpm@9.13.1",
   "private": true,
   "scripts": {
     "dev": "vite",
--- a/src/pages/AccountDetails.vue
+++ b/src/pages/AccountDetails.vue
@@ -350,7 +350,7 @@ const shouldShowInstructions = computed(() => transactionsTab.value === 'instruc
     }
 
     @include lg {
-      width: 46vw;
+      width: 45vw;
     }
 
     @include xl {
@@ -481,7 +481,7 @@ const shouldShowInstructions = computed(() => transactionsTab.value === 'instruc
       width: 90vw;
     }
     @include lg {
-      width: 46vw;
+      width: 48vw;
     }
     @include xl {
       width: size(80);
@@ -513,6 +513,35 @@ const shouldShowInstructions = computed(() => transactionsTab.value === 'instruc
         }
         & > .content-row {
           height: auto;
+        }
+
+        .transactions-table__columns {
+          @include sm {
+            flex-direction: column;
+          }
+          @include md {
+            flex-direction: row;
+          }
+          @include lg {
+            flex-direction: column;
+          }
+        }
+
+        .transactions-table__column {
+          &-hash {
+            @include sm {
+              width: 30vw;
+            }
+            @include md {
+              width: 19vw;
+            }
+            @include lg {
+              width: size(24);
+            }
+          }
+          &-block {
+            width: size(16);
+          }
         }
       }
     }
--- a/src/pages/AccountDetails.vue
+++ b/src/pages/AccountDetails.vue
@@ -1,14 +1,11 @@
 <script setup lang="ts">
 import { useRouter } from 'vue-router';
-import { computed, onMounted, ref } from 'vue';
+import { computed, reactive, ref, watch } from 'vue';
 import * as http from '@/shared/api';
 import BaseContentBlock from '@/shared/ui/components/BaseContentBlock.vue';
 import DataField from '@/shared/ui/components/DataField.vue';
-import { useTable } from '@/shared/lib/table';
 import BaseTable from '@/shared/ui/components/BaseTable.vue';
 import BaseLoading from '@/shared/ui/components/BaseLoading.vue';
-import { useErrorHandlers } from '@/shared/ui/composables/useErrorHandlers';
-import type { Account } from '@/shared/api/schemas';
 import { AccountIdSchema } from '@/shared/api/schemas';
 import { parseMetadata } from '@/shared/ui/utils/json';
 import BaseTabs from '@/shared/ui/components/BaseTabs.vue';
@@ -19,48 +16,10 @@ import InstructionsTable from '@/shared/ui/components/InstructionsTable.vue';
 import BaseLink from '@/shared/ui/components/BaseLink.vue';
 import { XS_WINDOW_SIZE } from '@/shared/ui/consts';
 import { useWindowSize } from '@vueuse/core';
+import { useParamScope } from '@vue-kakuyaku/core';
+import { handleParamScope } from '@/shared/api/handle-param-scope';
 
 const router = useRouter();
-const { handleUnknownError } = useErrorHandlers();
-
-const accountId = computed(() => {
-  const id = router.currentRoute.value.params['id'];
-
-  return AccountIdSchema.parse(id);
-});
-
-const account = ref<Account | null>(null);
-const isFetchingAccount = ref(false);
-
-onMounted(async () => {
-  try {
-    isFetchingAccount.value = true;
-    account.value = await http.fetchAccount(accountId.value);
-
-    if (account.value) {
-      const ownedEntities = [];
-
-      if (account.value.owned_assets) ownedEntities.push(assetsTable.fetch({ owned_by: accountId.value.toString() }));
-      if (account.value.owned_domains) ownedEntities.push(domainsTable.fetch({ owned_by: accountId.value.toString() }));
-
-      if (!ownedEntities.length) return;
-
-      await Promise.all(ownedEntities);
-    }
-  } catch (e) {
-    handleUnknownError(e);
-  } finally {
-    isFetchingAccount.value = false;
-  }
-});
-
-const assetsTable = useTable(http.fetchAssets);
-const domainsTable = useTable(http.fetchDomains);
-
-const transactionsTab = ref<TabAccountTransactions>('transactions');
-
-const shouldShowInstructions = computed(() => transactionsTab.value === 'instructions');
-
 const { width } = useWindowSize();
 
 const hashType = computed(() => {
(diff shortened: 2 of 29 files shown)


VladislavPopovSR:
Now we have the following fetching approach:
1. Create reactive `listState` - our query params
1.1 Create fetching function e.g. if some params should be ommitted
1.2 Write watch where resetting page if necessary (e.g. when we not on the initial page and apply filters)
2. Create scope. Some key's might be conditional
3. Write and use computed variables to access scope data

VladislavPopovSR:
Please take a look. I had this jenkins error
Снимок экрана 2024-11-15 в 00 56 30

As I understood, it caused by discrepancy between my local pnpm version and package.json `packageManager` field. 
So after many times I finally installed `9.13.1` and changed package.json version to this. Even now in spite of the successfull jenkins check it gives incompatible lockfile file.
Снимок экрана 2024-11-15 в 00 43 38
Maybe you are familiar with this? Seems an error on my side

src/shared/api/handle-param-scope.ts
0x009922, line 1: This isn't directly related to `shared/api`. It is just a composable which does something with `vue-kakuyaku`, which does something with promises.

src/shared/api/handle-param-scope.ts
```diff
@@ -0,0 +1,34 @@
+import type { Reactive } from 'vue';
+import { computed, reactive } from 'vue';
+import { useErrorHandlers } from '@/shared/ui/composables/useErrorHandlers';
+import { useErrorRetry, useTask, wheneverRejected } from '@vue-kakuyaku/core';
+
+export function handleParamScope<T, K>(
```
0x009922, line 6: `handle` is not a conventional prefix for composables (which this function is). More over, this function doesn't set up a scope (doesn't rely on EffectScope API). It is a plain composable that sets up some reactive logic.

Consider other options:

- `useAsyncData`
- `setupAsyncData`

src/shared/api/schemas.ts
```diff
@@ -21,12 +21,10 @@ export interface Paginated<T> {
...
-const PaginationParams = z
-  .object({
-    page: z.number(),
-    per_page: z.number(),
-  })
-  .partial();
+const PaginationParams = z.object({
+  page: z.number(),
+  per_page: z.number(),
+});
```
0x009922, line 27: I see that this schema isn't used anywhere. Therefore, it is enough to define it just as an `interface`.

src/shared/api/handle-param-scope.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  return reactive({
+    isLoading: computed(() => state.pending),
+    data: computed<K | null>(() => {
+      const res = state.fulfilled?.value;
+
+      if (!res) return null;
+
+      return res;
+    }),
```
0x009922, line 32:
```suggestion
    data: computed<K | null>(() => state.fulfilled?.value),
```

src/shared/api/handle-param-scope.ts
```diff
@@ -0,0 +1,34 @@
+import type { Reactive } from 'vue';
+import { computed, reactive } from 'vue';
+import { useErrorHandlers } from '@/shared/ui/composables/useErrorHandlers';
+import { useErrorRetry, useTask, wheneverRejected } from '@vue-kakuyaku/core';
+
+export function handleParamScope<T, K>(
+  payload: Reactive<T>,
+  fn: (query: Reactive<T>) => Promise<K>,
```
0x009922, line 8:
```suggestion
  fn: () => Promise<K>,
```

You can include the payload in the closure naturally, without needing to introduce any arguments here:

```ts
handleParamScope(payload, http.fetchTransactions)
// vs
handleParamScope(() => http.fetchTransactions(payload))
```

src/shared/ui/components/BasePagination.vue
0x009922, line 1: Please move the complicated and messy logic into a separate module, and organise it as pure functions (input -> output, no side effects, just the logic itself), and cover them with tests. This way, the dirty staff will be localised in a single place, not spoiling the rest code with complexity and mess.

This must have been an initial approach with this code. TDD is very helpful in such a case.

src/shared/ui/components/InstructionsTable.vue
```diff
@@ -45,36 +44,50 @@ function getInstructionPayloadEntity(item: Instruction) {
...
+    });
+  } else
+    return await http.fetchInstructions({
+      ...objectOmit(params, shouldShowKind.value ? ['kind', 'transaction_status'] : ['transaction_status']),
+    });
+}
 
-watch(listState, fetchInstructions, { immediate: true });
+watch(
+  () => [listState.kind, listState.transaction_status, listState.per_page],
```
0x009922, line 72: This watch will trigger even if the actual values in the array aren't changed (because the array object will be new each time this function is called).

Make three separate watch sources instead:
```suggestion
  [() => listState.kind, () => listState.transaction_status, () => listState.per_page],
```

src/shared/ui/components/InstructionsTable.vue
```diff
@@ -45,36 +44,50 @@ function getInstructionPayloadEntity(item: Instruction) {
...
+  transaction_status: null,
+  authority: computed(() => (props.filterBy.kind === 'authority' ? props.filterBy.value.toString() : '')),
+  transaction_hash: computed(() => (props.filterBy.kind === 'transaction' ? props.filterBy.value : '')),
   kind: 'All' as ftm.TabInstructions,
+  page: 1,
+  per_page: 10,
 });
 
-const shouldShowKind = computed(() => listState.kind === 'All');
+async function fetchInstructions(params: Reactive<typeof listState>) {
```
0x009922, line 59: The `Reactive` type is probably misused. `listState`'s type is already with all refs unwrapped, which is what `reactive()` returns.

0x009922, line 59: I'd suggest removing this function.

Instead, compute the actual params reactively:

```ts
const params = computed<InstructionSearchParams>(() => ...)
```

and setup the scope with them directly:

```ts
useParamScope(
  () => ({ key: JSON.stringify(params.value), payload: params }),
  ({ payload }) => handleParamScope(() => http.fetchInstructions(payload))
)
```

This way it could turn out that the whole `listState` is extra.

src/shared/ui/components/TransactionsTable.vue
0x009922, line 1: Same comments as in `InstructionsTable`.

src/widgets/home-page-info/HomePageInfo.vue
```diff
@@ -99,8 +99,12 @@ const nodesAmount = ref(0);
 
 const { handleUnknownError } = useErrorHandlers();
 
+const isLoading = ref(false);
```
0x009922, line 102: Replace with `usePromise` or `useTask`.

The point of this PR is to reduce this repetitive manual data-fetching related reactivity logic.

src/widgets/latest-blocks/LatestBlocks.vue
```diff
@@ -47,33 +47,33 @@ import BaseLink from '@/shared/ui/components/BaseLink.vue';
...
+  () => {
+    return {
+      key: 'latest-blocks',
+      payload: {
+        per_page: 10,
+      },
+    };
+  },
+  ({ payload }) => handleParamScope(payload, http.fetchBlocks)
+);
```
0x009922, line 69: The point of setting up a scope is that you have independent virtual reactive environments for each set of parameters. Thus, if parameters change, you don't need to set up various watchers to stop current fetching and start a new one _manually_ - the scope is simply disposed (as if it was a separate component) and a new one is set up.

In this case payload is constant, thus it's pointless to set up a parametrised scope. It only introduces extra complexity.

src/widgets/latest-transactions/LatestTransactions.vue
```diff
@@ -62,54 +62,57 @@
...
-      isLoading.value = false;
-    }
+watch(
+  total,
+  () => {
+    emit('loaded', total.value);
   },
-  { immediate: true }
+  { once: true }
 );
```
0x009922, line 116: A more correct and elegant way would be to put it directly into the scope where data is loaded, and use the `wheneverFulfilled` helper. 

```ts
wheneverFulfilled(state, (value) => emit('loaded', value.pagination.total_items))
```

0x009922 (changes requested):

src/widgets/latest-transactions/LatestTransactions.vue
VladislavPopovSR: Thats great idea but I've just realized that there is no need to fetch transactions and blocks - we can use `blocks`, `txs_rejected` and `txs_accepted` fields when fetching peer status in order to get their total count

package.json
```diff
@@ -32,8 +33,9 @@
     "zod": "^3.23.8"
   },
   "devDependencies": {
+    "@stylistic/eslint-plugin-ts": "^2.11.0",
```
0x009922, line 36: Why?

src/shared/ui/composables/usePagination.ts
0x009922, line 1: It is still not what I meant.

Inside `usePagination`, you have computation of `segmentInfo` and of `numbers`. Make them separate functions which have nothing to do with reactivity, _just pure logic_:

```ts
function computeSegmentInfo(input: {
	totalItems?: number,
	page: number,
	reversed: boolean
	/* ... */
}): { start: number, end: number, total: number } 

function computeNumbers(input: {
	/* put your params here */
}): (number | 'ellipsis')[]  
```

Then, test these functions directly. Pure input, pure output. All corner cases.

Then, define a composable, `usePagination`, on top of it. Please take a look at how composables are usually defined in VueUse. Specifically, I don't like that you pass all parameters in a single `ComputedRef<...>`. Make them all separate, make each `MaybeRefOrGetter<T>`.

Then, on top of it, add a few high-level tests of the actual component `<BasePagination>`, which doesn't test pagination, but only how it is displayed.

src/shared/utils/setup-async-data.ts
```diff
@@ -0,0 +1,26 @@
...
+  useErrorRetry(state, run, { interval: options?.interval ?? 5000 });
+
+  wheneverRejected(state, (err) => {
+    handleUnknownError(err);
+  });
+
+  return reactive({
+    isLoading: computed(() => state.pending),
+    data: computed<K | undefined>(() => state.fulfilled?.value),
+  });
```
0x009922, line 25: Why not just return `{ state }` directly.

src/widgets/latest-blocks/LatestBlocks.vue
```diff
@@ -47,34 +47,14 @@ import BaseLink from '@/shared/ui/components/BaseLink.vue';
...
-import { onMounted, ref, shallowRef } from 'vue';
-import { useErrorHandlers } from '@/shared/ui/composables/useErrorHandlers';
+import { computed } from 'vue';
 import TimeStamp from '@/shared/ui/components/TimeStamp.vue';
+import { setupAsyncData } from '@/shared/utils/setup-async-data';
 
-const emit = defineEmits<{
-  loaded: [number]
-}>();
+const setup = setupAsyncData(() => http.fetchBlocks({ per_page: 10 }));
```
0x009922, line 54: Again, look at how composables are defined in VueUse. You usually can use them in a way like this:

```ts
const { isLoading, data: blocks } = setupAsyncData(...) 
``` 

Consider aligning your composable with what's an established standard in the ecosystem.

0x009922 (changes requested):

package.json
VladislavPopovSR: Well DevOps said that there was a problem with `typescript-eslint/eslint-plugin` package. After I've installed the latest version I've found out that some rules are deprecated/moved to new packages. E.g. [member-delimiter-style](https://typescript-eslint.io/rules/member-delimiter-style/) rule.

src/shared/utils/setup-async-data.ts
VladislavPopovSR: In order to access only `expose` field since we don't use other fields for now and every time have the same output with only two fields: `isLoading` and `data`

VladislavPopovSR:
I think for `BasePagination.test.ts` there is no need to move data to variable and reuse it since every test has different amount of cases and different fields passed within

package.json
0x009922: There is no longer a problem with packages in CI. Can we revert extra changes?

src/widgets/latest-transactions/LatestTransactions.vue
```diff
@@ -62,56 +62,46 @@
...
-      isLoading.value = true;
-
-      const params = { per_page: 5, status: status.value };
+const listState = reactive({
+  per_page: 5,
+  status: null,
+});
 
-      const res = await http.fetchTransactions(objectOmit(params, status.value ? [] : ['status']));
+async function fetchTransactions(params: Reactive<typeof listState>) {
```
0x009922, line 85: Why do you use `Reactive` here?

src/pages/TransactionDetails.vue
```diff
@@ -44,24 +42,24 @@ const transactionHash = computed(() => {
...
-    } finally {
-      isFetchingTransaction.value = false;
-    }
+const transactionScope = useParamScope(
+  () => {
+    return {
+      key: transactionHash.value,
+      payload: transactionHash.value,
+    };
   },
```
0x009922, line 51: Primitive payloads could be passed as-is (they themselves will be used as the key): ([docs](https://www.npmjs.com/package/@vue-kakuyaku/core#reactively-parametrised-scope-with-useparamscope)):

```suggestion
  transactionHash,
```

src/pages/DomainsList.vue
```diff
@@ -113,6 +99,29 @@ const hashType = computed(() => {
 
   return 'two-line';
 });
+
+const listState = reactive({
+  page: 1,
+  per_page: 10,
+});
+
+watch([() => listState.per_page], () => {
```
0x009922, line 108:
```suggestion
watch(() => listState.per_page, () => {
```

src/pages/AccountDetails.vue
```diff
@@ -19,55 +16,93 @@ import InstructionsTable from '@/shared/ui/components/InstructionsTable.vue';
...
+});
 
-const { width } = useWindowSize();
+const assetsScope = useParamScope(
+  () => {
+    return {
+      key: account.value?.owned_assets ? JSON.stringify(assetsListState) : '',
+      payload: assetsListState,
+    };
+  },
```
0x009922, line 96: Please read the documentation of `useParamScope`. It supports optional scopes:
```suggestion
  () => {
    return (account.value?.owned_assets || false) && {
      key: JSON.stringify(assetsListState),
      payload: assetsListState,
    };
  },
```

src/pages/BlockDetails.vue
```diff
@@ -30,31 +28,31 @@ const blockHeightOrHash = computed(() => {
...
+  ({ payload }) => setupAsyncData(() => http.fetchBlock(payload))
 );
 
+const isBlockLoading = computed(() => blockScope.value.expose.isLoading);
+const block = computed(() => {
+  const res = blockScope.value?.expose.data;
+
+  if (!res) return null;
+
+  return res;
```
0x009922, line 47:
```suggestion
  return blockScope.value?.expose.data ?? null;
```

src/pages/DomainDetails.vue
```diff
@@ -44,35 +42,70 @@ const domainId = computed(() => {
...
 
-const domain = ref<Domain | null>(null);
-const isFetchingDomain = ref(false);
+const domainScope = useParamScope(
+  () => {
+    return {
+      key: domainId.value,
+      payload: domainId.value,
+    };
+  },
```
0x009922, line 51: Primitive payload

src/pages/AccountDetails.vue
VladislavPopovSR: Great! I decided to use `if` condition instead and return either `null` or scope.

src/widgets/latest-transactions/LatestTransactions.vue
VladislavPopovSR: Forgot to remove :(

0x009922 (commented):
LGTM, but there are some pending conversations

0x009922 (approved):

package.json
VladislavPopovSR: Let's follow newer versions and get away from deprecated functionality

PR #71, opened 2025-03-02
Feat - Integrate iroha sdk

VladislavPopovSR:
- Integrated `@iroha/core` 
- Some ui improvements
- Removed old and unused `iroha2` package


src/shared/ui/components/BaseTable.vue
```diff
@@ -81,6 +83,13 @@ interface Props {
   reversed?: boolean
 }
 
+const emit = defineEmits<{
+  'click:row': [data: T]
+}>();
+
+const instance = getCurrentInstance();
+const hasClickRowHandler = computed(() => !!instance?.vnode?.props?.['onClick:row']);
```
0x009922, line 91: If it isn't _really_ necessary to access raw VNodes and other Vue internals, better not to.

The same could be achieved by `useAttrs()`, or by simply providing an explicit boolean prop like `row-pointer`.

vite.config.mts
```diff
@@ -3,6 +3,7 @@ import { defineConfig } from 'vite';
 import path from 'path';
 import vue from '@vitejs/plugin-vue';
 import svg from 'vite-svg-loader';
+import wasm from "vite-plugin-wasm";
```
0x009922, line 6: Btw, it might not work when building for production (didn't for me). Have you checked if `vite build` output works?

VladislavPopovSR: Well locally it builds well as usual. I've also tried `vite preview` - nothing wrong. 
`continuous-integration/jenkins/pr-merge` completed successfully so I guess there are no issues

0x009922 (approved):

vite.config.mts
0x009922: > I've also tried `vite preview` - nothing wrong.

Hm, interesting.

PR #73, opened 2025-03-04
Feat - Update master

VladislavPopovSR:
Update `master` to the latest stable version


0x009922 (commented):
Better to do so at the same time with the backend.

It still has a blocker that Iroha isn't published on crates.io.

VladislavPopovSR:
Develop branch will be merged instead when we reach stability.

PR #74, opened 2025-03-13
Feat - Update to latest sdk

VladislavPopovSR:


src/widgets/home-page-info/HomePageInfo.vue
```diff
@@ -14,17 +14,19 @@
...
-          v-else
-          class="home-page-info__item-value"
-        >
-          {{ item.value }}
-        </span>
+        <div class="home-page-info__item-content">
+          <BaseLoading
+            v-if="isLoading"
+            class="home-page-info__item-content-loading"
+          />
```
0x009922, line 21: I am concerned that having 6 spinners on the main page for each data point is somewhat excessive from the UI standpoint.

All this data is fetched in a single request. One loading process. 6 spinners.

Maybe just one spinner would be enough, or a "skeleton" placeholder.

VladislavPopovSR: Sure thing!

Снимок экрана 2025-03-14 в 16 34 34

0x009922 (approved):

PR #76, opened 2025-03-18
feat: add `serve-mock-metrics.ts`

0x009922:
This PR adds a simple server script that mocks backend implementation of https://github.com/hyperledger-iroha/iroha/issues/5348. The purpose is to streamline development of the frontend.

How to use:

1. Install latest Deno (v2+)
2. Uncomment line in `vite.config.mts` with `server.proxy` config for `/api/v1/metrics`
3. Run `pnpm dev:serve-mock-metrics`

Server API is as specified in https://github.com/hyperledger-iroha/iroha/issues/5348#issuecomment-2728221129:

>  * add `GET /metrics` - return `NetworkMetrics`, summary metrics
>  * add `GET /metrics/peers` - return `PeerMetrics[]`, the latest metrics from the peers. With `?sse` query flag, this endpoint will instead use [Server-sent events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) and continuously emit metrics once available.
>  * add `GET /metrics/peers/info` - return `PeerInfo[]`, metadata about peers such as their IDs, URLs, configuration parameters

### DIFF
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -8,8 +8,7 @@ import svg from 'vite-svg-loader';
 export default defineConfig({
   server: {
     proxy: {
-      // Enable when running `serve-mock-metrics.ts`
-      // '/api/v1/metrics': 'http://localhost:8099',
+      '/api/v1/metrics': 'http://localhost:8099',
       '/api/v1': 'http://127.0.0.1:4000',
     },
   },
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -8,6 +8,8 @@ import svg from 'vite-svg-loader';
 export default defineConfig({
   server: {
     proxy: {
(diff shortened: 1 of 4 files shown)


0x009922:
@VladislavPopovSR does it work locally for you?

vite.config.mts
```diff
@@ -8,6 +8,8 @@ import svg from 'vite-svg-loader';
 export default defineConfig({
   server: {
     proxy: {
+      // Enable when running `serve-mock-metrics.ts`
+      // '/api/v1/metrics': 'http://localhost:8099',
```
0x009922, line 12:
```suggestion
      '/api/v1/metrics': 'http://localhost:8099',
```

VladislavPopovSR (approved):

PR #77, opened 2025-03-24
Feat - Add telemetry page

VladislavPopovSR:
Also added tests for `countTimeDifference` function

Part of https://github.com/hyperledger-iroha/iroha/issues/5348

- [x] Open a separate ticket to optimise re-rendering of the page (https://github.com/hyperledger-iroha/iroha/issues/5348)


src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,353 @@
...
+<template>
+  <div class="nodes-telemetry-page">
+    <BaseLoading
+      v-if="isLoading"
+      class="nodes-telemetry-page_loading"
+    />
+    <div
+      v-else-if="metrics"
+      class="nodes-telemetry-page__stats"
+    >
```
0x009922, line 84: I notice full layout blink (disappearance and appearance) at the time of the re-fetch.

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,353 @@
...
+      <div class="nodes-telemetry-page__stats-stat">
+        <span class="nodes-telemetry-page__stats-stat-value">{{ Math.trunc(metrics.average_block_time_ms) / 1000 }}s</span>
+        <span class="nodes-telemetry-page__stats-stat-label">{{ $t('telemetry.averageBlockCommitTime') }}</span>
+      </div>
+      <div class="nodes-telemetry-page__stats-stat">
+        <span class="nodes-telemetry-page__stats-stat-value nodes-telemetry-page__stats-stat-last-block">
+          <span
+            class="nodes-telemetry-page__stats-stat-last-block-timespan"
+            :class="{ 'nodes-telemetry-page__stats-stat-last-block-timespan_big': Number(formattedLastBlock) >= 10 }"
+          >{{ formattedLastBlock }}</span>
```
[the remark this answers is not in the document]
VladislavPopovSR: Moved to separate component. Preview mode:
Снимок экрана 2025-03-26 в 19 34 22

src/shared/api/index.ts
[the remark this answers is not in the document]
VladislavPopovSR: Decided to use the following instead:
```
useEventSource('/api/v1/metrics/peers?sse', ['metrics'])
```

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,411 @@
...
+
+const metrics = ref<NetworkMetrics | null>(null);
+const peersInfo = shallowRef<PeerInfo[]>([]);
+const peers = ref<Peer[]>([]);
+
+const { isLoading, execute } = useAsyncState(fetchMetrics, null, {
+  immediate: false,
+});
+
+const { data, close, status } = useEventSource('/api/v1/metrics/peers?sse', ['metrics']);
```
0x009922, line 42: Please move direct interaction with the API into `api/index.ts` and expose it as a method so that `NodesTelemetry.vue` doesn't need to know about the exact API use and only receives data.

src/shared/api/schemas.ts
```diff
@@ -153,6 +153,64 @@ export const PeerStatus = z.object({
...
+    .nullable(),
+  queue_capacity: z.number().nullable(),
+  connected_peers: z.string().array().nullable(),
+});
+
+export type PeerInfo = z.infer<typeof PeerInfo>;
+
+const Peer = PeerMetrics.merge(PeerInfo);
+
+export type Peer = z.infer<typeof Peer>;
```
0x009922, line 212: Please remove this. Metrics and info are separate data entities, although they have the same shared key - `public_key`/`peer`.

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,411 @@
...
+
+onMounted(async () => {
+  peersInfo.value = await http.fetchPeersInfo();
+});
+
+async function fetchMetrics() {
+  try {
+    metrics.value = await http.fetchNetworkMetrics();
+  } catch (e) {
+    handleUnknownError(e);
```
0x009922, line 68: I believe there was some different error handling approach in other components. Please make it consistent.

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,411 @@
...
+}
+
+const { pause } = useIntervalFn(execute, 15000, {
+  immediateCallback: true,
+});
+
+onUnmounted(() => {
+  pause();
+  close();
+});
```
0x009922, line 79: Isn't it called automatically in these composables?

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,411 @@
...
+  return (Math.floor(lastBlockTimestamp.value.value / 100) / 10).toFixed(1);
+});
+const lastBlockTimestamp = computed(() => {
+  if (!metrics.value?.latest_block_created_at) return null;
+
+  return useTimeAgo(metrics.value.latest_block_created_at, {
+    refreshInterval: 100,
+    detailedSeconds: true,
+  });
+});
```
0x009922, line 93: Unused.

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,411 @@
...
+  if (width.value >= SM_WINDOW_SIZE) return 'medium';
+
+  if (width.value >= XS_WINDOW_SIZE) return 'short';
+
+  return 'short';
+});
+
+const metrics = ref<NetworkMetrics | null>(null);
+const peersInfo = shallowRef<PeerInfo[]>([]);
+const peers = ref<Peer[]>([]);
```
0x009922, line 36: This is not an efficient structure for this data. It is very natural to use `Map` for peers data, as they are indexed by their `public_key`.

```ts
type PeerData = {
  metrics: null | PeerMetrics
  info: null | PeerInfo
}

const peers = reactive(new Map<string, PeerData>())
```

- When receiving `PeerInfo[]` from `/metrics/peers/info`, it should create/update entries in this map, with `info` field. Note that on further calls to `/metrics/peers/info` there could be new peers, and some peers could disappear. So this update should cleanup entries in this map.
- When receiving `PeerMetrics` from `/metrics/peers?sse`, it should fill entries in this map too. It is possible that there will be some `PeerMetrics` for `peer` that wasn't returned in the latest `PeerInfo[]` set. Therefore, update of the map should potentially create new `PeerData` entries with `info: null`.
- Having it as a `Map` makes this updates fast and efficient - no need to traverse the whole list, remove/insert elements in it.

src/pages/NodesTelemetry.vue
0x009922: This screencast of the index page, not `NodesTelemetry.vue`.

src/shared/api/schemas.ts
```diff
@@ -153,6 +153,64 @@ export const PeerStatus = z.object({
...
+
+export const PeerMetrics = z.object({
+  time: z.string(),
+  peer: z.string(),
+  role: z.enum(['Leader', 'ProxyTail', 'ValidatingPeer', 'ObservingPeer']),
+  block: z.number(),
+  block_created_at: z
+    .string()
+    .nullable()
+    .transform((x) => (x ? new Date(x) : null)),
```
0x009922, line 182: Please use dates as described in [Zod docs](https://zod.dev/?id=dates-1).

VladislavPopovSR: Done in 1388525

src/pages/NodesTelemetry.vue
VladislavPopovSR: Changed to `Map` structure usage✅

src/pages/NodesTelemetry.vue
VladislavPopovSR: I didn't check it before but yeah, it does. Removed✅

src/pages/NodesTelemetry.vue
VladislavPopovSR: Changed to `useAsyncData`✅

src/shared/api/schemas.ts
VladislavPopovSR: Removed✅

src/pages/NodesTelemetry.vue
VladislavPopovSR: Sure! I guess the following should be fine. `events` argument can be added if needed
```
export function streamPeerMetrics() {
  return useEventSource('/api/v1/metrics/peers?sse', ['metrics']);
}
```

src/shared/api/index.ts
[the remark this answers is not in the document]
VladislavPopovSR: I don't think there is a need of generator function while `useEventSource` composable provides all the necessary features

src/pages/NodesTelemetry.vue
VladislavPopovSR: I thought you were talking about blinking as in the video above.
Here are telemetry page's state screenshots: 
Снимок экрана 2025-03-27 в 18 33 33
Снимок экрана 2025-03-27 в 18 37 52
Снимок экрана 2025-03-27 в 18 38 25
Снимок экрана 2025-03-27 в 18 43 25
The gap between pictures 3 and 4 happens once in 10 times and quickly (while the metrics are loaded and the info  not yet and on the contrary).
I guess you mean `BaseContentBlock` height expansion? Should I apply `BaseLoading` on it instead of `BaseTable` then? Overall, currenly it looks fine even with 3G throttling.

src/pages/NodesTelemetry.vue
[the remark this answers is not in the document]
VladislavPopovSR: Actually, here the result without latest block component. I suggest working on chunks, performance and optimization in another pull request within separated task. And yeah, as you've already mentioned, it's essential topic🤝
Снимок экрана 2025-03-28 в 00 39 07
Снимок экрана 2025-03-28 в 00 41 20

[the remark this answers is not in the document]
0x009922: > I suggest working on chunks, performance and optimization in another pull request

Good suggestion, agreed.

src/pages/NodesTelemetry.vue
0x009922: This is better, but not complete. `JSON.parse` is still called externally and casted to the actual data. It is better if this representation detail is hidden within the `api` module.

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,397 @@
...
+    });
+  },
+});
+
+const { data: streamedPeerMetrics, status: streamStatus } = streamPeerMetrics();
+
+watch(streamedPeerMetrics, () => {
+  if (!streamedPeerMetrics.value || !peersInfo.value) return;
+
+  const peerMetrics = PeerMetrics.parse(JSON.parse(streamedPeerMetrics.value));
```
0x009922, line 65: This parsing should happen in `streamPeerMetrics()`, with returned `data` being `Ref<PeerMetrics>` or something.

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,397 @@
...
+    data?.forEach((i) => {
+      peers.set(i.public_key, { info: { ...i }, metrics: null });
+    });
+  },
+});
+
+const { data: streamedPeerMetrics, status: streamStatus } = streamPeerMetrics();
+
+watch(streamedPeerMetrics, () => {
+  if (!streamedPeerMetrics.value || !peersInfo.value) return;
```
0x009922, line 63: This is not correct.  

```suggestion
  if (!streamedPeerMetrics.value) return;
```

Metrics could arrive earlier. On the first connection, they arrive for all peers. What happens for me locally is that data is completely disregarded until `info` is loaded.

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,397 @@
...
+  onError: handleUnknownError,
+  resetOnExecute: false,
+});
+const { state: peersInfo } = useAsyncState(http.fetchPeersInfo, null, {
+  immediate: true,
+  onError: handleUnknownError,
+  onSuccess: (data) => {
+    data?.forEach((i) => {
+      peers.set(i.public_key, { info: { ...i }, metrics: null });
+    });
```
0x009922, line 56: This is not correct.

```suggestion
    invariant(data)
    data.forEach((i) => {
      peers.set(i.public_key, { ...(peers.get(i.public_key) ?? { metrics: null }), info: i });
    });
```

1. Info could arrive after metrics. 
2. Info should be re-fetched once in a while (e.g. 30 seconds), and it must keep existing metrics data
3. There should no be case when `data` is undefined

0x009922 (changes requested):

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,397 @@
...
+watch(streamedPeerMetrics, () => {
+  if (!streamedPeerMetrics.value || !peersInfo.value) return;
+
+  const peerMetrics = PeerMetrics.parse(JSON.parse(streamedPeerMetrics.value));
+  const peer = peers.get(peerMetrics.peer);
+
+  peers.set(peerMetrics.peer, { info: peer?.info ?? null, metrics: peerMetrics });
+});
+
+useIntervalFn(getNetworkMetrics, 15000, {
```
0x009922, line 71: This timeout is too large, but let's keep it for now. I will most probably make it so that `/metrics` are also streamed via SSE.

src/pages/NodesTelemetry.vue
0x009922: I no longer observe this behaviour.

I think it is because previously whenever the new data was loading, old data was cleared, thus creating `v-else-if` to be false.

Now you use `useAsyncState`, which seems to keep old data till the new is loaded. My speculation.

0x009922:
Any ideas why this warning happens?

```
Deprecation Warning [mixed-decls]: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls

   ┌──> src/pages/NodesTelemetry.vue
48 │           color: theme-color('content-primary');
   │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ declaration
   ╵
   ┌──> src/shared/ui/styles/breakpoints.scss
22 │ ┌   @media (min-width: $lg) {
23 │ │     @content;
24 │ │   }
   │ └─── nested rule
   ╵
    src/pages/NodesTelemetry.vue 48:9  root stylesheet

```

src/pages/NodesTelemetry.vue
VladislavPopovSR: Moved to api layer✅

src/pages/NodesTelemetry.vue
VladislavPopovSR: Moved into api layerRemoved✅

src/pages/NodesTelemetry.vue
VladislavPopovSR: Removed✅

src/pages/NodesTelemetry.vue
VladislavPopovSR: Formatted and added refetch every 30s✅

src/pages/NodesTelemetry.vue
VladislavPopovSR: Sure

VladislavPopovSR:
> Any ideas why this warning happens?
> 
> ```
> Deprecation Warning [mixed-decls]: Sass's behavior for declarations that appear after nested
> rules will be changing to match the behavior specified by CSS in an upcoming
> version. To keep the existing behavior, move the declaration above the nested
> rule. To opt into the new behavior, wrap the declaration in `& {}`.
> 
> More info: https://sass-lang.com/d/mixed-decls
> 
>    ┌──> src/pages/NodesTelemetry.vue
> 48 │           color: theme-color('content-primary');
>    │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ declaration
>    ╵
>    ┌──> src/shared/ui/styles/breakpoints.scss
> 22 │ ┌   @media (min-width: $lg) {
> 23 │ │     @content;
> 24 │ │   }
>    │ └─── nested rule
>    ╵
>     src/pages/NodesTelemetry.vue 48:9  root stylesheet
> ```


According to official [documentation](https://sass-lang.com/documentation/breaking-changes/mixed-decls/#deprecating-the-old-way), newer `sass` version has many breaking changes. Regarding this one, they want us to move some styles upper than other declarations. I've done that to every page, now there is no such warning✅

There are also a few other warnings left:
- Deprecation Warning [import]: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.
- Deprecation Warning [global-builtin]: Global built-in functions are deprecated and will be removed in Dart Sass 3.0.0.
Use map.has-key instead.

I will take a deeper look at it later since it covers entire project in order not to destroy current UI

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,383 @@
+<script setup lang="ts">
+import * as http from '@/shared/api';
+import BaseTable from '@/shared/ui/components/BaseTable.vue';
+import BaseContentBlock from '@/shared/ui/components/BaseContentBlock.vue';
+import { computed, reactive, watch } from 'vue';
+import { formatNumber } from '@/shared/ui/utils/formatters';
+import type { PeerInfo } from '@/shared/api/schemas';
+import type { PeerMetrics } from '@/shared/api/schemas';
```
0x009922, line 8: Sonar seems to produce a useful warning here

> '@/shared/api/schemas' imported multiple times.

src/shared/api/index.ts
```diff
@@ -89,6 +94,24 @@ export async function fetchPeerStatus(): Promise<PeerStatus> {
...
 
+export async function fetchNetworkMetrics(): Promise<NetworkMetrics> {
+  const res = await get('/metrics');
+  return NetworkMetrics.parse(res);
+}
+
+export function streamPeerMetrics() {
+  const { data: streamedPeerMetrics, status } = useEventSource('/api/v1/metrics/peers?sse', ['metrics']);
+  return {
+    data: computed(() => (streamedPeerMetrics.value ? PeerMetrics.parse(JSON.parse(streamedPeerMetrics.value)) : null)),
```
0x009922, line 105: Are you sure that this `computed` re-computes for every single change of `data`?

It is possible that there will be **a lot of events coming almost simultaneously**, especially on the first connection. I am not sure whether using `computed` will not cause it "seeing" only the last one within a single "tick".

(generally, for this specific streaming endpoint, I think there is no benefit of using a reactivity for data object, and a callback approach would be more robust and simple. No computeds, no watches, just `onData: (data) => ...`)

I checked and it seems to work fine though... What do you think is better to do?

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,383 @@
...
+      class="nodes-telemetry-page__stats"
+    >
+      <div class="nodes-telemetry-page__stats-stat">
+        <span class="nodes-telemetry-page__stats-stat-value">#{{ formatNumber(metrics.latest_block) }}</span>
+        <span class="nodes-telemetry-page__stats-stat-label">{{ $t('telemetry.bestBlock') }}</span>
+      </div>
+      <div class="nodes-telemetry-page__stats-stat">
+        <span class="nodes-telemetry-page__stats-stat-value">#{{ formatNumber(metrics.finalized_block) }}</span>
+        <span class="nodes-telemetry-page__stats-stat-label">{{ $t('telemetry.finalizedBlock') }}</span>
+      </div>
```
0x009922, line 106: These could be links to the block details page.

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,383 @@
...
+        :loading="streamStatus === 'CONNECTING'"
+        :items="Array.from(peers.values())"
+        container-class="nodes-telemetry-page__list-container"
+        :breakpoint="1440"
+      >
+        <template #header>
+          <div class="nodes-telemetry-page__list-row">
+            <span class="h-sm cell">{{ $t('telemetry.publicKey') }}</span>
+            <span class="h-sm cell">{{ $t('telemetry.publicUrl') }}</span>
+            <span class="h-sm cell">{{ $t('telemetry.block') }}</span>
```
0x009922, line 143: This could also be a link to the block details page.

src/pages/NodesTelemetry.vue
```diff
@@ -0,0 +1,383 @@
...
+              >{{ item.info.location ?? 'Unknown' }}</span>
+            </div>
+          </div>
+        </template>
+      </BaseTable>
+    </BaseContentBlock>
+  </div>
+</template>
+
+<style lang="scss">
```
0x009922, line 252: image

Inconsistency: monospace fonts are used for the stats, while numbers in the table are not monospaced.

0x009922: I would also suggest using it for public keys.

Maybe we shall create a separate task to create all tables in Explorer consistent and using monospace fonts where it is useful.

VladislavPopovSR: Edited to `monospace`. Will talk to designer about it 👍

src/pages/NodesTelemetry.vue
VladislavPopovSR: Added✅

src/pages/NodesTelemetry.vue
VladislavPopovSR: Fixed✅

src/shared/api/index.ts
0x009922: If there is no problem, then it's fine.

`onData` callback would be something simple like:

```ts
// in component
streamPeersMetrics({
  onData(data: PeerMetrics) {
    // update reactive state with the new data
  }
})
```

src/pages/NodesTelemetry.vue
0x009922: I see that you added it in the header, but not in the table.

I mean something like:
```suggestion
            <span class="h-sm cell"><BaseLink :to="...">{{ $t('telemetry.block') }}</BaseLink></span>
```

src/shared/api/index.ts
VladislavPopovSR: Alright then, let's merge

src/pages/NodesTelemetry.vue
0x009922: My bad - this is the table's header...

0x009922 (approved):

PR #83, opened 2025-04-10
refactor: remove store assets, remove `/status` usage

VladislavPopovSR:


0x009922 (approved):
The parts related to metrics seems to be unfinished, but so does the part of the backend. LGTM.

PR #93, opened 2025-05-14
build: Dockerfile - build & serve with Caddy

0x009922:


Dockerfile
```diff
@@ -1,3 +1,21 @@
...
+COPY pnpm-lock.yaml package.json .
+RUN pnpm fetch
+
+COPY src src
+COPY public public
+COPY *.json *.ts *.cjs *.mts *.html .
+RUN pnpm install
+RUN pnpm build
+
+FROM caddy:2-alpine
```
BAStos525, line 16: Motivation to switch the base image? We should keep to use unprivileged rootless images

Dockerfile
```diff
@@ -1,3 +1,21 @@
...
+
+COPY Caddyfile /etc/caddy/Caddyfile
+COPY --from=builder /app/dist /srv
+
+# NOTE: Caddy will serve on port 80
```
BAStos525, line 21: For deployments, post 80 is taken either by traefik or ingess-nginx balancer. Is it possible to serve web server on 8080 port?

nginx.conf
```diff
@@ -1,30 +0,0 @@
-user  nginx;
-worker_processes  1;
```
BAStos525, line 2: Why to remove?

Dockerfile
0x009922: Sure, changed to 8080.

Dockerfile
0x009922: Switched to Caddy from NGINX in order to pass `BACKEND_HOST` and `BACKEND_PORT` envs for variable backend reverse proxy setup.

> We should keep to use unprivileged rootless images

Noted, applied changes - no Caddy runs rootless.

BAStos525 (approved):

BAStos525 (approved):

PR #97, opened 2025-05-21
Feat - Update telemetry

VladislavPopovSR:


0x009922:
> Additionally updated `zod` to the latest version ([zod.dev/v4](https://zod.dev/v4) - quite an incredible speed up).

Omg this is amazing!

src/shared/api/index.ts
```diff
@@ -101,20 +102,56 @@ export async function fetchBlock(heightOrHash: number | string): Promise<Block>
...
+          };
+        }
+        default: {
+          return {
+            kind: 'first',
+            data: PeerInitialMetrics.parse(receivedData),
+          };
+        }
+      }
+    }),
```
0x009922, line 148: Zod supports [discriminated unions](https://zod.dev/api?id=discriminated-unions), please use them.

src/shared/lib/localization/en.json
```diff
@@ -156,15 +157,25 @@
...
     "averageBlockCommitTime": "Average block commit time",
     "lastBlock": "Last block",
     "publicKey": "Public key",
-    "publicUrl": "Public URL"
+    "publicUrl": "Public URL",
+    "networkGossipDetails": "size: {0}\nperiod: {1}ms",
+    "blocksGossipStats": "{0} per {1}ms",
+    "connected": "Connected",
+    "disconnected": "Disconnected",
+    "peerDoesntProvideTelemetry": "this peer doesn't provide telemetry data yet"
```
0x009922, line 179:
```suggestion
    "peerDoesNotProvideTelemetry": "This peer does not provide telemetry data"
```

src/pages/BlockDetails.vue
```diff
@@ -86,6 +103,12 @@ const hashType = useAdaptiveHash({ xxl: 'full', xl: 'full' });
         >
           <BaseLoading />
         </div>
+        <div
+          v-else-if="isFetchingBlockFailed"
+          class="block-details__metrics_unavailable row-text"
+        >
+          {{ $t('blocks.blockNotAvailableYet') }}
+        </div>
```
0x009922, line 111: This is not strictly the correct condition to display this message.

This message should be displayed _only_ when backend returns error 404 for the given `GET /blocks/{height}` request.

src/pages/NodesTelemetry.vue
```diff
@@ -86,36 +76,45 @@ function formatTimeSpan(date1: Date | null, date2: Date | null) {
...
-        <span class="nodes-telemetry-page__stats-stat-value">{{ Math.trunc(metrics.average_commit_time_ms) / 1000 }}s</span>
+        <span class="nodes-telemetry-page__stats-stat-value">{{ Math.trunc(metrics.avg_block_time.ms) / 1000 }}s</span>
         <span class="nodes-telemetry-page__stats-stat-label">{{ $t('telemetry.averageBlockTime') }}</span>
       </div>
       <div class="nodes-telemetry-page__stats-stat">
-        <span class="nodes-telemetry-page__stats-stat-value">{{ Math.trunc(metrics.average_block_time_ms) / 1000 }}s</span>
+        <span
+          v-if="metrics.avg_commit_time"
+          class="nodes-telemetry-page__stats-stat-value"
+        >{{ Math.trunc(metrics.avg_commit_time.ms) / 1000 }}s</span>
```
0x009922, line 107:
```suggestion
        >{{ ~~(metrics.avg_commit_time.ms / 1000) }}s</span>
```

src/pages/NodesTelemetry.vue
```diff
@@ -138,121 +137,239 @@ function formatTimeSpan(date1: Date | null, date2: Date | null) {
...
             />
+
+            <span
+              class="row-text cell"
+              :class="[
+                item.info.connected
+                  ? 'nodes-telemetry-page__list-row-connection-success'
+                  : 'nodes-telemetry-page__list-row-connection-error',
+              ]"
+            >{{ item.info.connected ? $t('telemetry.connected') : $t('telemetry.disconnected') }}</span>
```
0x009922, line 183: How about replacing "Connected"/"Disconnected" word with a simple green/red dot with a text tooltip?

src/pages/NodesTelemetry.vue
```diff
@@ -138,121 +137,239 @@ function formatTimeSpan(date1: Date | null, date2: Date | null) {
...
+              type="medium"
             />
+
+            <span
+              class="row-text cell"
+              :class="[
+                item.info.connected
+                  ? 'nodes-telemetry-page__list-row-connection-success'
+                  : 'nodes-telemetry-page__list-row-connection-error',
+              ]"
```
0x009922, line 182:
```suggestion
              :data-connected="item.info.connected"
```

This is less verbose. Then select in CSS by e.g. `[data-connected="true"]`

src/shared/api/index.ts
VladislavPopovSR: Done✅
Actually I didn't forget about such feature, I've been thinking before that it only works with specifying fields like the following:

```ts
z.object({kind:'...', field1, field2})
```
but actually it works with spread operator meaning 👍 
```ts
PeerInfo.extend({ kind: z.literal('peer_info') })
```

src/pages/NodesTelemetry.vue
VladislavPopovSR: Done✅

src/pages/NodesTelemetry.vue
VladislavPopovSR: Decided to use colored dot the same as we have on `BlocksList` page:

Снимок экрана 2025-05-23 в 19 24 32
Снимок экрана 2025-05-23 в 19 25 01

src/pages/BlockDetails.vue
VladislavPopovSR: Actually I've just realised that there is no way to filter such errors since every time we receive the same error - plain text. Maybe we could specify these errors on the backend side? 
e.g. if block height or hash is valid - send certain message
otherwise - error 404 not found

src/pages/BlockDetails.vue
```diff
@@ -31,7 +31,7 @@ const blockHeightOrHash = computed(() => {
 const isFetchingBlockFailed = ref(false);
 
 function handleBlockFetchingError(err: unknown) {
-  if (err instanceof ApiError && err.status === 404) isFetchingBlockFailed.value = true;
+  if (err === NOT_FOUND_ERROR) isFetchingBlockFailed.value = true;
```
0x009922, line 34: AFAIU this won't work: you are throwing `new Error(NOT_FOUND_ERROR)` (`Error` object with message = `NOT_FOUND_ERROR`), but here you check whether the whole error `=== NOT_FOUND_ERROR`).

VladislavPopovSR: Actually we receive plain string here and it works properly, but yeah lets use strict unions instead

src/shared/api/index.ts
```diff
@@ -95,26 +108,48 @@ export async function fetchBlocks(params?: Partial<PaginationParams>): Promise<P
...
 
-export async function fetchBlock(heightOrHash: number | string): Promise<Block> {
-  const res = await get(`/blocks/${heightOrHash}`);
-  return Block.parse(res);
+export async function fetchBlock(
+  heightOrHash: number | string
+): Promise<
+  | { status: typeof SUCCESS_FETCHING_BLOCK_STATUS, data: Block }
+  | { status: typeof BLOCK_NOT_FOUND_STATUS }
+  | { status: typeof ERROR_FETCHING_BLOCK_STATUS }
```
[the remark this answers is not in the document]
VladislavPopovSR: Added
```ts
type ResultWithStatus<T> = { status: SuccessFetchingStatus, data: T } | ErrorResponse;

export type ErrorResponse = { status: NotFoundStatus } | { status: UnknownErrorStatus, error: Error };
```

src/shared/api/index.ts
```diff
@@ -37,6 +43,13 @@ async function get<T>(path: string, params?: Record<string, any>): Promise<T> {
   }
 
   const res = await fetch(url);
+
+  if (!res.ok) {
+    if (res.status === 404) throw new Error(NOT_FOUND_ERROR);
+
+    throw new Error(await res.text());
+  }
```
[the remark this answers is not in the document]
VladislavPopovSR: Changed API requests to the following approach:
```ts
export async function fetchBlock(heightOrHash: number | string): Promise<ResultWithStatus<Block>> {
  const res = await get(`/blocks/${heightOrHash}`);
  if (res.status === SUCCESS_FETCHING_STATUS) return { status: SUCCESS_FETCHING_STATUS, data: Block.parse(res.data) };

  return await transformErrorResponse(res.response);
}
```

src/pages/BlockDetails.vue
```diff
@@ -25,15 +32,43 @@ const blockHeightOrHash = computed(() => {
...
 
 const isBlockLoading = computed(() => blockScope.value.expose.isLoading);
-const block = computed(() => blockScope.value?.expose.data);
+const block = computed(() => {
+  if (blockScope.value?.expose.data?.status === SUCCESS_FETCHING_BLOCK_STATUS) return blockScope.value.expose.data.data;
+
+  return null;
+});
+const isBlockNotFound = computed(() => blockScope.value?.expose.data?.status === BLOCK_NOT_FOUND_STATUS);
+const isOtherError = computed(() => blockScope.value?.expose.data?.status === ERROR_FETCHING_BLOCK_STATUS);
```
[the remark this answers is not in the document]
0x009922: My suggestion is fully type safe. Adding consts introduces verbosity which seems extra to me. 

> I strongly insist to reuse const values

Why?

